PR #3, opened 2020-07-16
Styled components + front migration [CU-6mtkw8]

Fuzznimp:


package.json
```diff
@@ -16,23 +16,43 @@
     "graphql-tag": "2.10.4",
     "next": "9.4.4",
     "react": "16.13.1",
-    "react-dom": "16.13.1"
+    "react-dom": "16.13.1",
+    "react-feather": "2.0.8",
```
Fenntasy, line 20: Just putting a comment that came to mind but it's not blocking for this PR.
I'm not against `react-feather` in principle but in the future, it would be great to get rid of it.

Here is my thinking: we need to manage a repo that will be forked by other people that will add their design systems on top of it. It would then be great to not "impose" a dependency that these other system would use.

But then again, this PR is already big so let's not do it now ^^

src/command/addIdentityToUser.ts
```diff
@@ -0,0 +1,35 @@
...
+    $type: IdentityTypes!
+    $value: String!
+  ) {
+    addIdentityToUser(input: { userId: $userId, type: $type, value: $value }) {
+      id
+      primary
+      status
+      type
+      value
+      __typename
```
Fenntasy, line 19: curiosity, why is that needed?

src/components/InputForm.tsx
```diff
@@ -0,0 +1,60 @@
+import React from "react";
+import styled from "styled-components";
+
+import { IdentityTypes } from "../@types/Identity";
+
+export const InputForm: React.FC<{ type: IdentityTypes }> = ({ type }) => {
+  const [formValue, setFormValue] = React.useState("");
```
Fenntasy, line 7: This is more of a `EmailIdentityForm` with `const [email, setEmail] = React.useState("")` don't you think?

src/design-system/theme/useTheme.ts
```diff
@@ -0,0 +1,9 @@
+import { useContext } from "react";
+import { ThemeContext } from "styled-components";
+
+import { LightTheme } from "./lightTheme";
```
Fenntasy, line 4: Thinking out loud, this would be a good use case for having a default export in a `theme.ts` file.
We could check in it a Config and return either the light theme as a default or the theme set in the config.
Having a default export would allow us to not name it since we wouldn't know which one is selected

src/pages/api/identity.tsx
```diff
@@ -0,0 +1,44 @@
...
+      })
+      .catch((error) => {
+        console.log(error);
+
+        response.statusCode = 400;
+
+        return response.json("failed");
+      });
+  }
+  response.statusCode = 400;
```
Fenntasy, line 41: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405

And can we do something like `response.status(405).end()`?

src/pages/api/identity.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return response.json(":(");
```
Fenntasy, line 43:
```suggestion
  return response.end();
```

src/pages/api/identity.tsx
```diff
@@ -0,0 +1,44 @@
...
+        response.statusCode = 200;
+
+        return response.json("succeed");
+      })
+      .catch((error) => {
+        console.log(error);
+
+        response.statusCode = 400;
+
+        return response.json("failed");
```
Fenntasy, line 38:
```suggestion
        return response.json({error: "Could not remove identity"});
```

src/pages/api/identity.tsx
```diff
@@ -0,0 +1,44 @@
...
```
Fenntasy, line 24:
```suggestion
        return response.json({error: "Could not add identity"});
```

src/pages/api/identity.tsx
```diff
@@ -0,0 +1,44 @@
...
+  response: NextApiResponse,
+): Promise<void> => {
+  const { userId, type, value } = request.body;
+
+  if (request.method === "POST") {
+    return await addIdentityToUser({ userId, type, value })
+      .then(() => {
+        response.statusCode = 200;
+
+        return response.json("succeed");
```
Fenntasy, line 17: The custom is usually to return either the added thing (and that could help you don't need to refresh the page)

src/components/InputForm.tsx
Fuzznimp: Yeah true, I wanted to have a global component but it's not needed there. I'll change it to make it more readable.

src/command/addIdentityToUser.ts
Fuzznimp: For now, none of these are need (even the ID, as we don't use apollo caching). Do you want me to remove them?

src/pages/api/identity.tsx
Fuzznimp: Good error code to know. I can't use status as a method though, it's typed as number.

```ts
    class ServerResponse extends OutgoingMessage {
        statusCode: number;
        statusMessage: string;

        constructor(req: IncomingMessage);

        assignSocket(socket: Socket): void;
        detachSocket(socket: Socket): void;
        // https://github.com/nodejs/node/blob/master/test/parallel/test-http-write-callbacks.js#L53
        // no args in writeContinue callback
        writeContinue(callback?: () => void): void;
        writeHead(statusCode: number, reasonPhrase?: string, headers?: OutgoingHttpHeaders): this;
        writeHead(statusCode: number, headers?: OutgoingHttpHeaders): this;
        writeProcessing(): void;
    }
```

Fuzznimp: So I'll just change the error code (the error message has been changed in the next pr, so I won't change it here)

src/pages/api/identity.tsx
Fuzznimp: Ok nvm, I'll change them there.

src/pages/api/identity.tsx
Fuzznimp: Quick question, as the `response` object is massive. Do I have the data added in the `response` object ? Or do I, if succeed, return what was in the payload?

Fuzznimp: I.e.
```ts
return await addIdentityToUser({ userId, type, value })
      .then((data) => {
        response.statusCode = 200;

        return response.json({data});
      })
```

Fenntasy: I think you may need some adjustments on that, depending on what is the return. 

Take care also that errors in a graphQL request are sent as a success in HTTP so I think we may need to check if there are errors in there

Fuzznimp: > Take care also that errors in a graphQL request are sent as a success in HTTP so I think we may need to check if there are errors in there 

Yeah I realised that while building the app. Do you want me to make the change here? Maybe a new task would be wiser.

Fenntasy: Go for a new task

src/design-system/theme/useTheme.ts
Fuzznimp: I see the idea, but I think we should talk about it to be sure I get the full picture.

Fuzznimp: Made the task

src/components/EmailIdentityForm.tsx
```diff
@@ -3,7 +3,9 @@ import styled from "styled-components";
 
 import { IdentityTypes } from "../@types/Identity";
 
-export const InputForm: React.FC<{ type: IdentityTypes }> = ({ type }) => {
+export const EmailIdentityForm: React.FC<{ type: IdentityTypes }> = ({
+  type,
+}) => {
   const [formValue, setFormValue] = React.useState("");
```
Fenntasy, line 9: Well, if you made the change here, what about 

`const [email, setEmail] = React.useState("")` ? 😇

Fuzznimp: Woops 🙇

Fuzznimp: Wait, actually, this is also used to add/remove phone number 🤦‍♂️. Sorry, I'm not awake.

```ts
<Flex>
            <Button
              onClick={() => setAddPhone(!addEmail)}
              color={theme.colors.green}
            >
              Add a phone number
            </Button>
            {addPhone && <EmailIdentityForm type={IdentityTypes.PHONE} />}
          </Flex>
```

Fuzznimp: f6daa83

src/components/IdentityInputForm.tsx
```diff
@@ -3,19 +3,19 @@ import styled from "styled-components";
 
 import { IdentityTypes } from "../@types/Identity";
 
-export const EmailIdentityForm: React.FC<{ type: IdentityTypes }> = ({
+export const IdentityInputForm: React.FC<{ type: IdentityTypes }> = ({
```
Fenntasy, line 6: 👍

PR #4, opened 2020-07-20
Add test to the index's `getServerSideProps` [CU-6mtm37]

Fuzznimp:


package.json
```diff
@@ -7,7 +7,8 @@
     "dev": "next dev -p $CONNECT_ACCOUNT_PORT",
     "build": "next build",
     "lint": "eslint \"{src,pages}/**/*.{js,ts,jsx,tsx}\"",
-    "start": "next start -p $CONNECT_ACCOUNT_PORT"
+    "start": "next start -p $CONNECT_ACCOUNT_PORT",
```
Fenntasy, line 10: I assume this is ill-advised if we want to deploy with vercel

package.json
```diff
@@ -61,5 +95,26 @@
...
+    ],
+    "testPathIgnorePatterns": [
+      "<rootDir>[/\\\\](node_modules|.next)[/\\\\]"
+    ],
+    "transformIgnorePatterns": [
+      "[/\\\\]node_modules[/\\\\].+\\.(ts|tsx)$"
+    ],
+    "transform": {
+      "^.+\\.(ts|tsx)$": "babel-jest"
+    }
```
Fenntasy, line 118: Starting to look like something I would prefer into its own config file 😅 
I see some similarities with what I can see there https://github.com/vercel/next.js/tree/canary/examples/with-jest
Do you think there are some things to take from them or did you go with the minimal settings?

Fuzznimp: I used this one https://github.com/vercel/next.js/blob/canary/examples/with-typescript-eslint-jest/jest.config.js as a baseline.

Regarding 
```
"moduleFileExtensions": [
      "js",
      "ts",
      "tsx",
      "json"
    ],
```

I left `js` and `json` in case we want to test `next.config.js` (dunno if there are any interest to do so), or the `config.json` in case we wan't to use this over var env.

I think the ignore patterns can be removed, like the `roots` field.

Fuzznimp: 794d75b
Let me know if it's short enough, or if I should also get ride of the `js` and `json` item.

package.json
Fuzznimp: Checking the vercel deploy config file doc atm https://vercel.com/docs/configuration#introduction/configuration-reference, not sure how it's handled (probably injecting env vars).

Fuzznimp: So the solution would be to have something like `"start": "next start -p $CONNECT_ACCOUNT_BUILD_PORT",`, and set up the env on the vercel console when we'll deploy, right ?

Fuzznimp: c8257f4

package.json
```diff
@@ -7,7 +7,8 @@
     "dev": "next dev -p $CONNECT_ACCOUNT_PORT",
     "build": "next build",
     "lint": "eslint \"{src,pages}/**/*.{js,ts,jsx,tsx}\"",
-    "start": "next start -p $CONNECT_ACCOUNT_PORT"
+    "start": "next start -p $CONNECT_ACCOUNT_BUILD_PORT",
```
Fenntasy, line 10: I remember we talked about that and in my mind we ended up with the default `start` port, am I misremembering?

package.json
```diff
@@ -21,38 +22,71 @@
...
-      [
-        "styled-components",
-        {
-          "ssr": true,
-          "displayName": true
-        }
-      ]
-    ]
+    "env": {
+      "development": {
```
Fenntasy, line 52: Should we bring back the `production` in this PR or let it in the next?

tests/serverSideProps.test.ts
```diff
@@ -0,0 +1,60 @@
...
+describe("getServerSideProps", () => {
+  beforeEach(() => {
+    fetch.resetMocks();
+  });
+
+  const mockedContext = {
+    req: new IncomingMessage(new Socket()),
+    res: new ServerResponse(new IncomingMessage(new Socket())),
+    query: {},
+  };
```
Fenntasy, line 19:
```suggestion
  const mockedRequest = new IncomingMessage(new Socket())
  const mockedContext = {
    req: mockedRequest,
    res: new ServerResponse(mockedRequest),
    query: {},
  };
```

package.json
Fuzznimp: We did, but I guessed that vercel injecting the port env var during build time would still require to specify said variable in the build and start script (with the same name as the one defined in the vercel dashboard). Am I wrong?

tests/serverSideProps.test.ts
Fuzznimp: That's weird.. that's the first thing I did when I mocked the request, but I had ts freaked out about it. I probably made a mistake.

package.json
Fuzznimp: I will add it as I need to do some more changes here.

PR #5, opened 2020-07-24
[CU-6rr09x] Implement first test with `testing-library`

Fuzznimp:


tests/__mocks__/fileMock.ts
```diff
@@ -0,0 +1,7 @@
+// This file, along the `moduleNameMapper` field found in the jest config,
+// is meant to fix the jest issue regarding assets imports during React component testing.
+//
+// For more information, please refer to this thread:
+// https://stackoverflow.com/questions/46898638/importing-images-breaks-jest-test
+
+export default "";
```
Fenntasy, line 7: I... alright

tests/addEmailIdentityWorkflow.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const { getByText, getByRole, getByTestId } = render(
+      <Index fetchedData={mockedFetchedData} />,
+      {},
+    );
+
+    const addAnEmailButton = getByText("Add an email");
+
+    fireEvent.click(addAnEmailButton);
+
+    const identityForm = getByTestId("identity-form");
```
Fenntasy, line 56: Not a blocking comment but are testid required? Could we envision another way or is it *really* simpler?

tests/addEmailIdentityWorkflow.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const emailInput = getByRole("textbox");
+
+    fireEvent.change(emailInput, { target: { value: "test@test.test" } });
+
+    const sendEmailButton = getByText("Send");
+
+    fireEvent.click(sendEmailButton);
+
+    expect(submitCheck).toBe(true);
```
Fenntasy, line 75: I have a little fear that this could be a flacky test in the future since we trigger a click that would trigger an event that could be asynchronous. Are you sure it always work?

tests/config/test-utils.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { lightTheme } from "../../src/design-system/theme/lightTheme";
+
+const AllTheProviders: React.FC<JSX.Element> = ({ children }) => {
+  return (
+    <ThemeProvider theme={lightTheme}>
+      <GlobalStyle />
+      <Layout>{children}</Layout>
+    </ThemeProvider>
+  );
+};
```
Fenntasy, line 17: Could that be done with the global App component?

tests/__mocks__/fileMock.ts
Fuzznimp: Yeah.. This makes no sense but that's the only thing that fixed the issue (hence the comment because this is really confusing).

tests/addEmailIdentityWorkflow.test.tsx
Fuzznimp: The only other option is to use something like `querySelector`, but we lose the goodies from `testing-library` if we do this. I really don't like the idea of changing components to enable testing like this, but that's the cleaner way I found.

For some reason, `form` is not a role so I can't select it with the lib. Maybe I could do it with catching the main content a searching for the child, or something along those line, but it would add a lot of bloat, and I wanted to go "fast".

tests/config/test-utils.tsx
Fuzznimp: Do you mean testing the app component rather than the index page ?

tests/addEmailIdentityWorkflow.test.tsx
Fenntasy: ok, well let's go with that at first then

tests/config/test-utils.tsx
Fenntasy: No, I mean not copy/pasting the providers but using the App component instead to avoid duplication and forgetting to add things there

tests/addEmailIdentityWorkflow.test.tsx
Fuzznimp: I am not, I'll spend more time on this playing with async processes.

tests/config/test-utils.tsx
Fuzznimp: I'm not sure I understand what you mean.

Fuzznimp: I mean, I understand the issue, but not what you have in mind as a solution.

Fenntasy:
Task linked: [CU-6rr09x Implement `testing-library` for user interaction testing.](https://app.clickup.com/t/6rr09x)

tests/addEmailIdentityWorkflow.test.tsx
Fuzznimp: I tested with adding an async api call on `https://api.spacexdata.com/v4/launches/latest` on submit, and logging its result. Everything works fine.

Worst case, we would need to use [their async helper](https://testing-library.com/docs/dom-testing-library/api-async) if this stop working. 

Let me know if you think it's good enough or not.

Fuzznimp: As seen together, it is working with async actions.

PR #9, opened 2020-07-27
Setup gh actions [CU-6utwrt]

Fuzznimp:
I think the deps install can be optimised, but I couldn't make it installed only one time for the 2 steps.


Fenntasy:
Task linked: [CU-6utwrt Add tests via Github Actions](https://app.clickup.com/t/6utwrt)

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,29 @@
...
+    
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: dependencies
+      run: yarn install
```
Fenntasy, line 14: Can we have cache? 😁

Fuzznimp: I guess this works cfbf96b 😅.

Also, @julien-leclercq, does adding the condition `if: steps.cache-primes.outputs.cache-hit != 'true'` will produce the issue you were talking this morning ? i.e having to add, as a first job
```yaml
success:
    name: "success"
    runs-on: ubuntu-latest
    steps:
      - name: "success"
        run: echo "success"
```

Would be great if you could explain real quick the deal to @Fenntasy.

PR #10, opened 2020-07-29
Add logger to ssr and api pages

Fuzznimp:


Fenntasy:
Task linked: [CU-62kk1f Setup Logging](https://app.clickup.com/t/62kk1f)

src/middleware/withAPIPageLogger.ts
```diff
@@ -0,0 +1,56 @@
+import { createLogger } from "@fewlines/fwl-logging";
+import { NextApiRequest, NextApiResponse } from "next";
+import { Handler } from "src/pages/api/identity";
+
+type WithAPIPageLogger = (
+  request: NextApiRequest,
+  response: NextApiResponse,
+) => Promise<JSON | void>;
```
Fuzznimp, line 8: Not sure if this should be type as `Promise<JSON | Error>` and I throw an Error instead of console logging it.

src/middleware/withAPIPageLogger.ts
```diff
@@ -0,0 +1,56 @@
...
+          path: request.url,
+          duration: (
+            (processTimeEnd - processTimeStart) /
+            BigInt(1000)
+          ).toString(),
+        });
+
+        response.statusCode = 400;
+
+        return response.json({ error: "Could not perform action" });
```
Fuzznimp, line 53: Not a fan of this, but we'll handle this with fwl I guess ?

src/middleware/withAPIPageLogger.ts
Fenntasy: Let me see if I understand correctly, this type is the return type of `WithAPIPageLogger` so the function that will be in the end passed to next for the API page? 

Shouldn't it be always `void` in that case since we don't return via the function but via the `response.json`?

src/middleware/withAPIPageLogger.ts
```diff
@@ -0,0 +1,56 @@
...
+        const processTimeEnd = process.hrtime.bigint();
+
+        logger.log("Success", {
+          method: request.method,
+          statusCode: request.statusCode,
+          path: request.url,
+          duration: (
+            (processTimeEnd - processTimeStart) /
+            BigInt(1000)
+          ).toString(),
```
Fenntasy, line 29: We didn't need `.toString` in FWL, is it needed there?

src/middleware/withAPIPageLogger.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function withAPIPageLogger(handler: Handler): WithAPIPageLogger {
+  const logger = createLogger("connect-account", "json").withMeta({
+    process: "getServerSideProps",
+  });
+
+  const processTimeStart = process.hrtime.bigint();
+
+  return async (request, response) => {
+    handler(request, response)
```
Fenntasy, line 18: Thinking out loud: we could have a problem there with passing `response`, inside this function, we could do `response.json(thing)` and it would result in an error of the type "response has already been sent".

Do you think it would be alright not to pass `response`?

src/middleware/withSSRLogger.ts
```diff
@@ -0,0 +1,51 @@
...
+    process: "getServerSideProps",
+  });
+
+  const processTimeStart = process.hrtime.bigint();
+
+  return async (context) => {
+    return getServerSideProps(context)
+      .then(
+        (
+          response: GetServerSidePropsResult<{
```
Fenntasy, line 18: is this a `response` or is it `props`?

src/middleware/withAPIPageLogger.ts
```diff
@@ -0,0 +1,56 @@
...
+          method: request.method,
+          statusCode: request.statusCode,
+          path: request.url,
+          duration: (
+            (processTimeEnd - processTimeStart) /
+            BigInt(1000)
+          ).toString(),
+        });
+
+        response.statusCode = 200;
```
Fenntasy, line 32: With this method, we can't send anything other than 200, which could pose a problem in the future

src/middleware/withAPIPageLogger.ts
Fuzznimp: We do, otherwise I get and error saying smth along the line of `Do not know how to serialize a BigInt` (I can reproduce the error if you want). We just need to add in the `log` function a type check like `typeof value === 'bigint' ? value.toString(): value`. I can write a task about it if you want.

src/middleware/withAPIPageLogger.ts
Fuzznimp: Yeah, I went with the simplest approach as I'm still new to the `statusCode` handling game, and as I don't know the "good" implementation, I just went for this

src/middleware/withSSRLogger.ts
Fuzznimp: It is `props`, but I went with this as it is the response from the promise above all imo (I was a bit confused about the semantic about this one). It is still working, but I can change the naming if you want.

src/middleware/withAPIPageLogger.ts
Fuzznimp: It is, but I remember you saying that returning `response.json` should be typed as `JSON` and not `void` (both are working).

src/middleware/withAPIPageLogger.ts
Fuzznimp: I see the issue (and got it during one of my first iteration for this task if I remember correctly).

I not sure about how to handle it this way thought. Would it means that we need to do something like `export default withAPIPageLogger(handler).then(({data}) => response.json({data})).catch(...);` ? Seems off.

src/middleware/withSSRLogger.ts
Fenntasy: well, seeing the type, what about `serverSidePropsResult`? A bit verbose but at least, we know 😁

src/middleware/withAPIPageLogger.ts
Fenntasy: Well, I would try something like:

```ts
interface HandlerReponse {
  status: HttpStatus;
  data?: unknown;
  headers?: Record<string, string>;
}
type Handler = (request: NextApiRequest) => Promise<HandlerResponse>
```

I'm not sure it covers all posibilities but I assume it fixes the http 200 problem along with the problem

src/middleware/withAPIPageLogger.ts
Fenntasy: I'm curious as to why there is no bug in FWL, is there and we just don't know it?

Fuzznimp: There are no tests that try to log a `bigInt` value. New task I guess ?

Fenntasy: ah yes, I had to fix it in FWL too ^^', my bad

src/middleware/withAPIPageLogger.ts
Fuzznimp: `withAPIPageLogger` shouldn't return a promise, right?

Fenntasy: Well, it can be either `void` or `Promise<void>` I guess, since you can export an async function

src/middleware/withAPIPageLogger.ts
```diff
@@ -0,0 +1,51 @@
...
+          path: request.url,
+          duration: Number((processTimeEnd - processTimeStart) / BigInt(1000)),
+        });
+
+        response.statusCode = HttpStatus.OK;
+      })
+      .catch((error: Error) => {
+        const processTimeEnd = process.hrtime.bigint();
+
+        console.log(error);
```
Fenntasy, line 36: oh hello there 👋 
Shouldn't it be a lint error?

Fuzznimp: Nop, I even ran the linter manually before committing. It was intentional, but now that I think about it, it's dumb as we have a logger.. I guess the console.log rule isn't in our config.

src/@types/ApiPageHandler.ts
```diff
@@ -0,0 +1,10 @@
+import { HttpStatus } from "@fewlines/fwl-web";
+import { NextApiRequest } from "next";
+
+interface HandlerResponse {
+  status: HttpStatus;
+  data?: unknown;
+  headers?: Record<string, string>;
```
Fenntasy, line 7: so, either we don't put it in this PR and we'll add it when needed or we deal with it in `withAPIPageLogger` ^^

src/middleware/withAPIPageLogger.ts
Fuzznimp: b58aa34

src/@types/ApiPageHandler.ts
Fuzznimp: Ohhhh damn I forgot about it..

Fuzznimp: The only way I have to deal with it, is to pass the header to the handler, right?

PR #11, opened 2020-07-29
Cu 6pn3rk fetch utils

Hamza-A:
I have created a new function custom fetch which can handle redundant code in each fetch request (setting the header and stringify the body), it will also lighten fetching components.


Fenntasy:
Task linked: [CU-6pn3rk [chore] Utils function that forward `fetch`](https://app.clickup.com/t/6pn3rk)

src/utils/customFetch.ts
```diff
@@ -0,0 +1,19 @@
+import { HttpVerbs } from "src/@types/HttpVerbs";
+
+type Body = {
+  [key: string]: string;
+};
```
Fuzznimp, line 5: As discussed, if you use the global Body type, you can remove this.

src/utils/customFetch.ts
```diff
@@ -0,0 +1,15 @@
+import { HttpVerbs } from "src/@types/HttpVerbs";
+
+export function customFetch(
+  endpoint: string,
+  method: HttpVerbs,
+  body: Record<string, unknown>,
```
Hamza-A, line 6: Following Anthony's advices, I'm using Record type instead of Body

src/utils/customFetch.ts
Hamza-A: OK I'm following your advices, I remove those lines and use type Record instead

src/utils/customFetch.ts
Fuzznimp: My bad, actually it is `Record<string, string | number>` 🙇

Hamza-A: OK fixed!

Fuzznimp: 💪

src/utils/customFetch.ts
```diff
@@ -0,0 +1,15 @@
+import { HttpVerbs } from "src/@types/HttpVerbs";
+
+export function customFetch(
```
Fenntasy, line 3: What do you think of naming it `fetchJson`?

"custom" does not convey any meaning ^^

src/utils/customFetch.ts
```diff
@@ -0,0 +1,15 @@
...
+  endpoint: string,
+  method: HttpVerbs,
```
Fenntasy, line 5: We should restrict that to the methods that are not `GET` since we won't be able to send a body with a GET

src/utils/customFetch.ts
Hamza-A: Sounds good, actually, at first I had named it jsonFetch before naming it custom fetch, I'm still hesitating

src/utils/customFetch.ts
Fuzznimp: Hamza, you could try to use [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html) to keep the enum (i.e. `Exclude`). I never used them on enums, but I don't see a reason why it wouldn't work.

src/utils/customFetch.ts
Hamza-A: OK fixed!

src/utils/customFetch.ts
Hamza-A: OK fixed!

Fuzznimp: One trick you can do, is copy the commit id, and post it in a comment, like this a8f2ae1.

src/utils/fetchJson.ts
```diff
@@ -1,6 +1,6 @@
 import { HttpVerbs } from "src/@types/HttpVerbs";
 
-type Method = Exclude<HttpVerbs, "GET">;
+type Method = Exclude<HttpVerbs, HttpVerbs.GET>;
```
Fuzznimp, line 3: I'd say `Methods` would precise that there are more than one. Also we may find a better wording, maybe `BodyMethods`? @Fenntasy Not sure.

Hamza-A: OK, for now it's changed for Methods 2d71ba8

Fenntasy: Well I'd stick with Method as we type the value `fetchJson` is expecting and it only expects one at a time.
It could be `HttpVerb` or `HttpMethod` but it's not linked to the body per se (the fact that there is a body is because of the method but the reverse is not true)

Fuzznimp: Make sense. What about `Method` vs `Methods`?

Fenntasy: Well let's reverse the thinking, you want to use `fetchJson` on VSCode, you type `fetchJson(` what do you expect to see in the argument lists? 

I would expect that the function wants a `Method`, wouldn't you?

Fuzznimp: I'd want to pick a method from a pool of method.

Fenntasy: yes but `Method` can be `POST | PUT | PATCH | DELETE` 😅

Fuzznimp: Then, you can revert to `Method` hamza, sorry 🙇.

src/utils/fetchJson.ts
```diff
@@ -1,10 +1,10 @@
 import { HttpVerbs } from "src/@types/HttpVerbs";
 
-export type Methods = Exclude<HttpVerbs, HttpVerbs.GET>;
+export type Method = Exclude<HttpVerbs, HttpVerbs.GET>;
```
Fuzznimp, line 3: As we can't merge you PR because of gh actions, just a comment:
If you need to export a type (i.e. it is used in more than one place), you should put the type in its own file, under `@types` 🙂.

Hamza-A: OK done! c7a43df

Fuzznimp: Pro 💪

PR #12, opened 2020-07-30
Cu 6mv5x7 sort identities

Hamza-A:
I added some logical in Index component to prevent the rendering of the delete button if the list of IdentityTypes (phone or email) is < 2.

I'm wrapping the FetchIconButton with an if statement and I'm adding new counters (emailCount and phoneCount), the delete button will render accordingly


Fenntasy:
Task linked: [CU-6mv5x7 Create function to sort identities to help render workflow](https://app.clickup.com/t/6mv5x7)

Fuzznimp:
Alright, couple points. I should have been more precise in describing the task, my bad 🙇.

The solution you propose works, but if you think in terms of scaling, this kind of implementation becomes quickly a nightmare to maintain. By doing this, you also make it harder to test the logic behind the conditional rendering of the delete button. You'd have to test the whole component, instead of just a function that manipulate the data. Also, we try to have the least possible amount of logic in rendered component. They are just there to show stuff based on conditions, not compute said condition.

So what I have in mind is a function that manipulate/format the identities array. So basically, giving the function `Identities[]`, and returning `(PhoneIdentities | MailIdentities | etc.)[][]` (let me know if you have issues understanding the TS syntax). This function should also be in its own file, under `utils`.

Doing this will have multiple benefices:
- You can (and you should in this task) test the function using `jest`. Identities are **central** to the application, so this should be **heavily** tested! Feel free to check the [jest documentation](https://jestjs.io/docs/en/getting-started), and ask questions!
- Formatting the data received will make the development process faster. Think of it as I'm giving you a messy tool box vs a well-organized one.

Now come the question, at which point do we manipulate the data? On the server side? On the client? Obviously, this is the job of the server to do this. If it would my call, I would format the data in the management's GraphQL resolver. As we don't have access to it, we have two choices left where to use it:
- `getIdentities`
- the index's `getServerSideProps`

For now, using it in one or the other does not change anything, as the data is fetch only in one place: the `index` page.

Doing it in `getIdentities` would remove the need to reapply the formatting every time we want to get the data in a new place. But as we still don't know the final architecture of the app, I would personally use it in the index `getServerSideProps`, and refactor it in the future if needed.

Then again, it's my fault for not being precise enough. I'm still glad you found a way of doing it quickly, even if it was a naive approach 🙂.

src/pages/index.tsx
```diff
@@ -24,6 +24,17 @@ const Index: React.FC<IndexProps> = ({ fetchedData }) => {
...
+  let emailCount = 0;
+  let phoneCount = 0;
+
+  user.identities.map(({ type }) => {
+    if (type === IdentityTypes.EMAIL.toLowerCase()) {
+      emailCount++;
+    } else if (type === IdentityTypes.PHONE.toLowerCase()) {
+      phoneCount++;
+    }
+  });
```
Fuzznimp, line 36: Also, `.map()` returns an array, which is not used here. If we would go for this implementation, `forEach()` would have been more efficient (as it returns nothing).

Hamza-A: OK fixed 79a61d5

src/pages/index.tsx
```diff
@@ -100,11 +99,20 @@ const Index: React.FC<IndexProps> = ({ fetchedData }) => {
...
+  const sortedIdentities = await getIdentities().then((result) => {
+    if (result instanceof Error) {
+      throw new Error();
+    }
+    return sortIdentities(result);
+  });
+
+  if (Object.values(sortedIdentities).length === 0) {
+    throw new Error("No identities found");
+  }
```
Fuzznimp, line 111: @Fenntasy I'd like your thoughts on this one. This is the quick solution we found to handle the case where the identities field received in `sortIdentities` is null or undefined, and returns an empty object.

Fenntasy: Do we need to throw errors though? 

Couldn't we return `{ phoneIdentities: [], emailIdentities: [] }` and just display "No emails" and "No phones" when the array is empty in the UI?

Hamza-A: OK fixed 1217fa4

src/pages/index.tsx
```diff
@@ -100,17 +106,20 @@ const Index: React.FC<IndexProps> = ({ fetchedData }) => {
...
+    }
+    return sortIdentities(result);
+  });
+
+  return {
+    props: {
+      sortedIdentities,
+    },
+  };
+};
```
Fuzznimp, line 122: This should be wrapped with `withSSRLogger`:

```ts
export const getServerSideProps: GetServerSideProps = withSSRLogger(
  async () => {
    const sortedIdentities = await getIdentities().then((result) => {
    if (result instanceof Error) {
      throw new Error(result);
    }

    return sortIdentities(result);
  });

  return {
    props: {
      sortedIdentities,
    },
  };
);
```

Also, you can throw the actual error (i.e. `throw new Error(result);`)

src/utils/sortIdentities.ts
```diff
@@ -0,0 +1,26 @@
...
+      } else if (identity.type === IdentityTypes.PHONE.toLowerCase()) {
+        phoneIdentities.push(identity);
+      }
+    });
+
+    return { phoneIdentities, emailIdentities };
+  }
+
+  return { phoneIdentities: [], emailIdentities: [] };
+}
```
Fuzznimp, line 26:
```suggestion
export function sortIdentities(fetchedData: {
  data?: { provider: ProviderUser } | null;
}): SortedIdentities {
  if (fetchedData.data) {
    const identities: Identity[] = fetchedData.data.provider.user.identities;
    const phoneIdentities: Identity[] = [];
    const emailIdentities: Identity[] = [];

    identities.forEach((identity) => {
      if (identity.type === IdentityTypes.EMAIL.toLowerCase()) {
        emailIdentities.push(identity);
      } else if (identity.type === IdentityTypes.PHONE.toLowerCase()) {
        phoneIdentities.push(identity);
      }
    });
  }

  return { phoneIdentities, emailIdentities };
}
```

You can go with this, as `phoneIdentities` and `emailIdentities` are initialized with an empty array. If there are, for example, no phone identities, `phoneIdentities` won't be populated and will return an empty array 😁.

package.json
```diff
@@ -24,28 +24,27 @@
...
+    "eslint-plugin-prettier": "^3.1.4",
+    "eslint-plugin-react": "^7.20.5",
     "file-loader": "6.0.0",
     "jest": "^26.1.0",
     "jest-fetch-mock": "^3.0.3",
-    "prettier": "2.0.5",
+    "prettier": "^2.0.5",
     "ts-jest": "^26.1.2",
     "ts-node": "8.10.2",
     "typescript": "3.9.6",
```
Fuzznimp, line 50: I have not told you yet, but we want full control over dependencies versions. For that, we remove the `^` from the version. Could you also remove them? Thanks! 😁

tests/addEmailIdentityWorkflow.test.tsx
```diff
@@ -2,50 +2,37 @@ import { enableFetchMocks } from "jest-fetch-mock";
...
-            },
-          ],
-        },
+  const mockedFetchedData: SortedIdentities = {
+    emailIdentities: [
+      {
+        id: "8f79dcc1-530b-4982-878d-33f0def6a7cf",
+        primary: true,
+        status: "validated",
+        type: IdentityTypes.EMAIL,
```
Fuzznimp, line 18: Weird that the test is passing, as this should be `ReceivedIdentityTypes.EMAIL` I think (same for phone).

src/pages/index.tsx
Hamza-A: OK done 629b6f1

src/utils/sortIdentities.ts
Hamza-A: OK fixed ee7d9ff

package.json
Hamza-A: OK I  have rolled back the package.json file to its previous version 18e6418

Fuzznimp: Thanks mate! I'll handle the dependencies bump (I have a recurring task for that, set to once a month I think?). Just keep in mind to remove the `^` when you add new dependencies :).

tests/addEmailIdentityWorkflow.test.tsx
Hamza-A: Maybe they were passing because of the `.toLowerCase()` ? AnywayI have changed the type to `ReceivedIdentityTypes.EMAIL`

Fuzznimp: Yeah, you are right, but I'm more curious of why TypeScript is not screaming (because `IdentityType.X.toString() !== ReceivedIdentityType.X`).

src/utils/sortIdentities.ts
```diff
@@ -0,0 +1,23 @@
...
+}): SortedIdentities {
+  const phoneIdentities: Identity[] = [];
+  const emailIdentities: Identity[] = [];
+  if (fetchedData.data) {
+    const identities: Identity[] = fetchedData.data.provider.user.identities;
+    identities.forEach((identity) => {
+      if (identity.type === IdentityTypes.EMAIL.toLowerCase()) {
+        emailIdentities.push(identity);
+      } else if (identity.type === IdentityTypes.PHONE.toLowerCase()) {
+        phoneIdentities.push(identity);
```
Fuzznimp, line 17: You can remove the lower casing now 😁.

Fuzznimp: I think this is the only place where we do this, but you can double-check just to be sure.

Hamza-A: I did double-check and they were no other place where IdentityTypes were lower cased, it's fixed for this file 074d6a9

Fuzznimp: Perfect! Good job mate!

src/pages/index.tsx
```diff
@@ -102,11 +109,16 @@ export default Index;
 
 export const getServerSideProps: GetServerSideProps = withSSRLogger(
   async () => {
-    const fetchedData = await getIdentities();
+    const sortedIdentities = await getIdentities().then((result) => {
+      if (result instanceof Error) {
+        throw new Error();
```
Fenntasy, line 114:
```suggestion
        throw new Error(result);
```

src/queries/getIdentities.ts
```diff
@@ -1,5 +1,6 @@
 import { FetchResult } from "apollo-link";
 import gql from "graphql-tag";
+import { ProviderUser } from "src/@types/ProviderUser";
```
Fenntasy, line 3:
```suggestion
import { ProviderUser } from "../@types/ProviderUser";
```

src/utils/sortIdentities.ts
```diff
@@ -0,0 +1,23 @@
+import { ProviderUser } from "src/@types/ProviderUser";
```
Fenntasy, line 1:
```suggestion
import { ProviderUser } from "../@types/ProviderUser";
```

src/pages/index.tsx
Hamza-A: OK done c8e792e, I had to stringify the result though, class Error was expecting an argument of type string not of type error

Fenntasy: Well, yes... since that's an instance of Error, we should just `throw result`, sorry 🙇‍♂️

src/queries/getIdentities.ts
Hamza-A: OK fixed ffa46c5

src/utils/sortIdentities.ts
Hamza-A: OK fixed ffa46c5

src/pages/index.tsx
Hamza-A: No prob! fixed 745ec61

PR #16, opened 2020-08-06
Fixed sc theme global type

Fuzznimp:
This PR aims at fixing the global theme to be able to use autocomplete.

@Fenntasy My Only issue with that solution is that it creates 2 instances of the `LigthTheme` object (in the global type def and in `lightTheme.ts`). Unfortunately, I can't do:

```ts
import "styled-components";

import { Spacing } from "./src/design-system/theme/lightTheme";

declare module "styled-components" {
  export interface DefaultTheme extends LigthTheme {}
}
```

as I get `An interface declaring no members is equivalent to its supertype.eslint@typescript-eslint/no-empty-interface`. Should I add an es-lint ignore?


Fenntasy:
Task linked: [CU-6au9er [chore] Fix theme autocomplete](https://app.clickup.com/t/6au9er)

Fuzznimp:
Ok I just tested

```ts
declare module "styled-components" {
  // eslint-disable-next-line @typescript-eslint/no-empty-interface
  export interface DefaultTheme extends LightTheme {}
}
```
And it's not working, so I guess I have no choices.

Fuzznimp:
I also tried
```ts
import { lightTheme } from "./src/design-system/theme/lightTheme";

type ThemeInterface = typeof lightTheme;

declare module "styled-components" {
  interface DefaultTheme extends ThemeInterface {}
};
```
but no success

yarn.lock
```diff
@@ -1084,10 +1084,10 @@
   resolved "https://registry.yarnpkg.com/@emotion/unitless/-/unitless-0.7.5.tgz#77211291c1900a700b8a78cfafda3160d76949ed"
   integrity sha512-OWORNpfjMsSSUBVrRBVGECkhWcULOAJz9ZW8uK9qgxD+87M7jHRcvh/A96XXNhXTLmKcoYSQtBEX7lHMO7YRwg==
 
-"@fewlines/eslint-config@^3.0.0":
-  version "3.0.0"
-  resolved "https://registry.yarnpkg.com/@fewlines/eslint-config/-/eslint-config-3.0.0.tgz#2d8c2be46b8bfec8222942b8ccc999d6d6d714f3"
-  integrity sha512-5J9UukKUA4phA1zAgZJIcVLv8Ps5kpNGAD3jtPqe/Ee2HhLN6AZTlThmgUIqIvm/0ZpDI7d/mrBPNcnt0nOuYA==
+"@fewlines/eslint-config@2.1.1":
```
Fenntasy, line 1087: hu?

src/@types/styled-component.d.ts
```diff
@@ -0,0 +1,75 @@
+import "styled-components";
+
+import { Spacing } from "../design-system/theme/lightTheme";
```
Fenntasy, line 3: Shouldn't we define it there now?

yarn.lock
Fuzznimp: There was a conflict with the master (which adds husky). I did not re-generate the yarn-lock, that's probably why.

src/@types/styled-component.d.ts
Fuzznimp: Nice catch!

PR #17, opened 2020-08-07
Cu 6mv5x7 sort identities

Hamza-A:
I wrote more tests for sortIdentities


Fenntasy:
Task linked: [CU-6mv5x7 Create function to sort identities to help render workflow](https://app.clickup.com/t/6mv5x7)

tests/sortIdentities.test.ts
```diff
@@ -1,11 +1,224 @@
...
+
+    const mockedSortedResponse = {
+      emailIdentities: [],
+      phoneIdentities: [],
+    };
+
+    const call = sortIdentities(mockedResponse);
+    expect(call).toStrictEqual(mockedSortedResponse);
+  });
+  it("should render properly if it receives only email", () => {
```
Fenntasy, line 120: Small histeria but each tests should be separated by a space 😇

tests/sortIdentities.test.ts
```diff
@@ -1,11 +1,224 @@
...
+              {
+                id: "7f8d168a-3f65-4636-9acb-7720a212680e",
+                primary: true,
+                status: "validated",
+                type: ReceivedIdentityTypes.EMAIL,
+                value: "test@fewlines.co",
+              },
+              {
+                id: "8f79dcc1-530b-4982-878d-33f0def6a7cf",
+                primary: true,
```
Fenntasy, line 138: To have data that respect the API, this couldn't happen, we will only ever have one primary identity per identity type.

tests/sortIdentities.test.ts
```diff
@@ -1,11 +1,224 @@
...
+          name: "Mocked Provider",
+          user: {
+            id: "299d268e-3e19-4486-9be7-29c539d241ac",
+            identities: [
+              {
+                id: "7f8d168a-3f65-4636-9acb-7720a212680e",
+                primary: true,
+                status: "validated",
+                type: ReceivedIdentityTypes.EMAIL,
+                value: "test@fewlines.co",
```
Fenntasy, line 134: We try to avoid having real or real looking emails so we should have only emails in `.local` or `.test`

Fenntasy:
Could you do the same tests but with the rendered page?

Meaning testing that for each version we only display the `delete` button when it's relevant

tests/sortIdentities.test.ts
Hamza-A: OK fixed 5c4f179

tests/sortIdentities.test.ts
Hamza-A: OK fixed 95e3cbf

tests/sortIdentities.test.ts
Hamza-A: OK fixed 11770b4

tests/pages/account.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+    const emailTrash = component.find({ type: IdentityTypes.EMAIL });
+    const phoneTrash = component.find({ type: IdentityTypes.PHONE });
+    const noPhone = component.find(".no-phone");
+    const noEmail = component.find(".no-email");
+
+    expect(emailTrash).toHaveLength(0);
+    expect(phoneTrash).toHaveLength(0);
+    expect(noPhone).toHaveLength(2);
+    expect(noEmail).toHaveLength(2);
```
Fenntasy, line 226: Why 2? Shouldn't there be one per type?

Hamza-A: When there is on element displayed, Enzyme acts as if the length is 2

Fenntasy: 🤔 what does it display if you console.log `noPhone`?

tests/pages/account.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+    const component = mount(
+      <ThemeProvider theme={lightTheme}>
+        <GlobalStyle />
+        <Layout>
+          <Account sortedIdentities={mockedSortedResponse} />
+        </Layout>
+      </ThemeProvider>,
+    );
+
+    const emailTrash = component.find({ type: IdentityTypes.EMAIL });
```
Fenntasy, line 218: Thinking about it, could we make a find more precise and only search the component we're looking for and not only one of its props? (I'm thinking searching both the component and its prop)

tests/pages/account.test.tsx
Hamza-A: the `console.log(noPhone);` returns `ReactWrapper {}`

tests/pages/account.test.tsx
Hamza-A: Unfortunately this is the most precise I could be, because giving the component an `id` or `className` throws the following error:
`Property 'className' does not exist on type 'IntrinsicAttributes & FetchIconButtonProps & { children?: ReactNode; }'.ts(2322)`

Fenntasy: What about something like `component.find(FetchIconButton).find({ type: IdentityTypes.EMAIL })`?

tests/pages/account.test.tsx
Fenntasy: Maybe with https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/debug.html

tests/pages/account.test.tsx
Hamza-A: OK fixed! 6697393

tests/pages/account.test.tsx
Hamza-A: OK fixed aa72665 I decided to use `.hostNodes()` method to filter down the result of `.find()` to HTML elements only.

Besides I decided to check if `emailTrash` 's behaviour is as expected if there are 3 identities (because so far I was only testing with 0, 1 or 2)

PR #18, opened 2020-08-13
Implement sentry

Fuzznimp:
This PR aims at implementing `sentry` error monitoring across the app.

I've tested every pages/components locally and manually, and used a spreadsheet to be sure I did not forget anything:

image

Testing it in a build environment is pretty tricky, as throwing errors in the code base prevent the build. The only test I did was to put wrong env vars values, and sentry was catching everything.

The `sentry.ts` utils is pretty self explanatory. The `configureReq` function is called when manually where you want to add more info from the request for the sentry report.

The `Breadcrumb` feature is called across the board, to provide event history leading to the error.

The error handling component was override to catch errors happening at the start of the app rendering (i.e. `_document`).

Also, `config.connectAccountPublicSentryDSN` isn't used anywhere, as we are using the env vars `NEXT_PUBLIC_SENTRY_DSN` right away to have it injected in GitHub Actions. I left it just to be sure the `env_vars` file has been sourced (otherwise, `config.ts` will crash the app).


Fenntasy:
Task linked: [CU-78n876 Setup Sentry](https://app.clickup.com/t/78n876)

src/utils/sentry.ts
```diff
@@ -0,0 +1,39 @@
...
+    dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
+    enabled: process.env.NODE_ENV !== "test",
+  });
+}
+
+Sentry.configureScope((scope) => {
+  scope.setTag("Node environment", process.env.NODE_ENV);
+});
+
+export const configureReq = (
```
Fuzznimp, line 18: Let me know if you guys want better naming.

src/pages/api/oauth/callback.ts
```diff
@@ -9,49 +9,59 @@ import { config } from "../../../config";
...
-    request.session.set("user-jwt", parsedResponse.access_token);
+      jwt.verify(parsedResponse.access_token, callback.client_secret);
 
-    await request.session.save();
+      request.session.set("user-jwt", parsedResponse.access_token);
 
-    response.writeHead(HttpStatus.MOVED_TEMPORARILY, {
-      Location: "/account",
-    });
+      await request.session.save();
```
Meyclem, line 48: From where the `.save()` is coming?

src/utils/sentry.ts
Meyclem: I'm ok with it.
Maybe `configureRequest` but it's just me being annoying 😁
(feel free to leave it like this if you prefer)

src/pages/api/oauth/callback.ts
Fuzznimp: From the session package.

Meyclem: 🙇

Fuzznimp:
> I'm ok with it.
> Maybe `configureRequest` but it's just me being annoying 😁
> (feel free to leave it like this if you prefer)

I'd like to add `sentry` to the name. Maybe `sentryRequestMonitoring` or `sentryRequestReport` would be better.

src/pages/api/account.ts
```diff
@@ -4,25 +4,35 @@ import { Handler } from "../../@types/ApiPageHandler";
 import { addIdentityToUser } from "../../command/addIdentityToUser";
 import { removeIdentityFromUser } from "../../command/removeIdentityFromUser";
 import { withAPIPageLogger } from "../../middleware/withAPIPageLogger";
+import Sentry, { configureReq } from "../../utils/sentry";
 
 const handler: Handler = async (request, response) => {
-  const { userId, type, value } = request.body;
+  configureReq(request);
```
Fenntasy, line 10: Shouldn't we make a middleware?

src/pages/index.tsx
```diff
@@ -43,6 +44,8 @@ export default Index;
 
 export const getServerSideProps: GetServerSideProps = withSSRLogger(
   async (context) => {
+    configureReq(context.req);
```
Fenntasy, line 47: Why is there no try/catch there?

src/config.ts
```diff
@@ -61,6 +63,12 @@ function handleEnvVars(): void {
     process.exit(1);
   }
   config.connectAccountSessionSalt = process.env.CONNECT_ACCOUNT_SESSION_SALT;
+
+  if (!process.env.NEXT_PUBLIC_SENTRY_DSN) {
+    console.log("Missing NEXT_PUBLIC_SENTRY_DSN");
+    process.exit(1);
+  }
+  config.connectAccountPublicSentryDSN = process.env.NEXT_PUBLIC_SENTRY_DSN;
```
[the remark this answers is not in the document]
Fuzznimp: Yeah, I was wondering this too.

src/utils/sentry.ts
```diff
@@ -0,0 +1,39 @@
...
+  const headers = Object.entries(request.headers).reduce(
+    (acc, [key, value]) => (key !== "cookie" ? { ...acc, [key]: value } : acc),
+    {},
+  );
+
+  Sentry.configureScope((scope) => {
+    scope.setTag("url", request.url ? request.url : "");
+    scope.setTag("method", request.method ? request.method : "");
+
+    if ("query" in request) {
```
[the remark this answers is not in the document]
Fuzznimp: This is a type checking, as `query` and `body` not present in the `IncomingMessage` type, but are in `NextApiRequest`.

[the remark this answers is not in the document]
Fuzznimp: This is like saying `if(request is ofType NextApiRequest)`, but TypeScript does not have a sexy way of doing it..

src/pages/api/account.ts
Fuzznimp: Probably, but I'd say we should work on the middleware composition before adding more middleware, don't you think?

The redirect middleware is also right around the corner.

I don't mind coding this middleware in this PR though.

src/pages/index.tsx
Fuzznimp: Because there is no way an error happens there, I think, as the function is only declaring an object, and passing it to the React component (_already regret those words_).

But you are right, calling `configureReq` has no effect there.

src/pages/api/account.ts
Fenntasy: well, let's make that the next PR then

Fuzznimp: Should I still refacto this into a middleware?

src/utils/sentry.ts
```diff
@@ -0,0 +1,39 @@
...
+      scope.setContext("query", request.query);
+      scope.setContext("body", request.body);
```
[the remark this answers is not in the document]
Fuzznimp: Hmm, I need to think about it (and probably discuss it just to be sure).

src/pages/api/account.ts
Fuzznimp: Telephone log: we'll wait to have the middleware composition before refactoring this into a middleware.

src/utils/sentry.ts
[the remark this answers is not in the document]
Fuzznimp: I'm not sure that it is correct, but if the callback api page throws when receiving the redirection from connect (with the `auth_code` as payload), sentry will catch the jwt, right? (from the query string or the body)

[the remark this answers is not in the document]
Fenntasy: from the query string no, since there will only ever be the authorization code. They would need to catch the data exchanged during a `fetch` call inside the api request. Which shouldn't be appearing, I guess

src/utils/sentry.ts
```diff
@@ -15,11 +15,14 @@ Sentry.configureScope((scope) => {
...
 });
 
-export const configureReq = (
+export const addRequestScopeToSentry = (
   request: NextApiRequest | IncomingMessage,
 ): void => {
   const headers = Object.entries(request.headers).reduce(
-    (acc, [key, value]) => (key !== "cookie" ? { ...acc, [key]: value } : acc),
+    (acc, [key, value]) =>
+      key.toLowerCase() !== ("cookie" || "authorization")
```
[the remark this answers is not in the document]
Fuzznimp: Wait, it should have changed on Friday.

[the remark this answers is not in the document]
Fuzznimp: Nvm, I miss-read your comment. Do you mean that `!== ("cookie" || "authorization")` won't work?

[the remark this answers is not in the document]
Fuzznimp: I couldn't reproduce a case where an `Authroization` field was provided.

I read about this and the only thing I can think of is casing, or prefix (i.e. `WWW` or `PROXY`).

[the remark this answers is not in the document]
Fenntasy: Here's what I mean

```
>  "cookie" !== ("cookie" || "authorization")
false
> "other" !== ("cookie" || "authorization")
true
>  "authorization" !== ("cookie" || "authorization")
true
> ("cookie" || "authorization")
'cookie'
```

It wouldn't work in the case of `authorization` since your `||` will always return `cookie`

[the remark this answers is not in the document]
Fuzznimp: Damn, nice catch. Fixed.

src/utils/sentry.ts
```diff
@@ -20,7 +20,7 @@ export const addRequestScopeToSentry = (
 ): void => {
   const headers = Object.entries(request.headers).reduce(
     (acc, [key, value]) =>
-      key.toLowerCase() !== ("cookie" || "authorization")
+      key.toLowerCase() !== "cookie" && key.toLowerCase() !== "authorization"
```
Fenntasy, line 23:
```suggestion
      !["cookie", "authorization"].includes(key.toLowerCase())
```

Fuzznimp: Damn that's beautiful, I had to make the change ahah. I was wondering, should I put this in a function and test it quickly? Looks important.

PR #21, opened 2020-08-18
Extract business add identity

Fuzznimp:


Fenntasy:
Task linked: [CU-7enjcq Refacto add identity](https://app.clickup.com/t/7enjcq)

src/components/business/AddIdentity.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  const addIdentity = (value: string): Promise<Response> => {
+    const body = {
+      userId: data.userId,
+      type,
+      value,
+    };
+
+    return fetchJson("/api/account", HttpVerbs.POST, body);
```
Fenntasy, line 33: Should we assume that it will always return json and already do `response.json()` there?
It's really a question, I have no prior opinion on this

src/components/business/AddIdentity.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface AddIdentityProps {
+  type: IdentityTypes;
+  children: (props: {
+    addIdentity: (value: string) => Promise<Response>;
+  }) => JSX.Element;
+}
+
+export const AddIdentity: React.FC<AddIdentityProps> = ({ type, children }) => {
+  const { data, error } = useCookies();
```
Fenntasy, line 16: If all our little component do that call, we may need to think about using context somewhere to avoid calling the API every time we instanciate a component (for now, I think that's acceptable but to have that in mind)

Fuzznimp: I agree with that. I'll write a task to keep this in mind.

Fuzznimp: For now, it works as an auth gate, but when we'll implement an auth middle, we'll be able to refactor this into a context that just provide the user ID.

src/components/business/AddIdentity.tsx
Fuzznimp: I guess it would be more flexible to not return a JSON, as we would be able to throw specific errors based on what kind of fetch we want to achieve. Kind of thinking out loud, is that the kind of interrogation you had in mind?

Fenntasy: Yes, that and, whose job is it to deal with errors?

Fuzznimp: Next internals I guess? Or do you want us to handle errors and fallback components ourselves?

Fenntasy: Imagine if we want to have flash messages that pop up when an error occurs, we would need to catch them (for small errors in my mind, if all goes to hell, then next to the rescue)

Fuzznimp: So we keep this as it is, right?

Fenntasy: yup, for now

Fenntasy (approved):

PR #23, opened 2020-08-20
Routing

Fuzznimp:


Fenntasy:
Task linked: [CU-7ev1ak Routing](https://app.clickup.com/t/7ev1ak)

src/utils/getUserIdentityFromId.ts
```diff
@@ -0,0 +1,23 @@
...
+export async function getUserIdentityFromId(
+  context: GetServerSidePropsContext,
+): Promise<void> {
+  const serverRouter: ServerRouter = (context.req as IncomingMessage & {
+    ServerRouter: ServerRouter;
+  }).ServerRouter;
+  console.log(serverRouter);
+
+  serverRouter && console.log(serverRouter.query.id);
+}
```
Fuzznimp, line 23: @Fenntasy I left this file to show you how I'll query management when we'll be able to filter identities with the ID, and not the value (you can refer to the conv in slack's `connect` channel).

Two things worries me:
- `ServerRoute` is not type out of the box, which worries me.
- When I log the object, it first logs as `undefined`, then as the actual object. So the correct type would be:
```ts
ServerRouter = (context.req as IncomingMessage & {
    ServerRouter?: ServerRouter;
  }
```

This is all the only implementation I can think of, and I'm really not a fan of this. I pinged you on the slack conv to see if I did not miss anything.

Fenntasy: I'm sorry, I don't understand what you mean there.

This file would be called by the future file `/account/logins/${email.type}/${email.id}`? 
I don't understand what is `serverRouter` to be honest

src/pages/account/logins/index.tsx
```diff
@@ -49,7 +50,11 @@ const Account: React.FC<AccountProps> = ({ sortedIdentities }) => {
               return (
                 <IdentityBox key={email.value}>
                   <Flex>
-                    <Value>{email.value}</Value>
+                    <Value>
+                      <Link href={`/account/logins/${email.type}/${email.id}`}>
```
Fenntasy, line 54:
```suggestion
                      <Link href="/account/logins/[type]/[id]" as={`/account/logins/${email.type}/${email.id}`}>
```

Fuzznimp:
> This file would be called by the future file /account/logins/${email.type}/${email.id}?
> I don't understand what is serverRouter to be honest

Yes, from `getServerSideProps`. 
`ServerRouter` is an object injected by NextJS inside `context.req`, which looks like this:
```
ServerRouter {
  route: '/account/logins/[type]/[id]',
  pathname: '/account/logins/[type]/[id]',
  query: { type: 'email', id: '***' },
  asPath: '/account/logins/email/***',
  basePath: '',
  events: undefined,
  isFallback: false
}
```

So I would use `ServerRouter` to get the identity id and fetch management to serve the client.

PR #26, opened 2020-09-01
Cu 7ev1pg identity actions

Hamza-A:
This branch aims to move the `AddIdentityInputForm` from the logins page to its own page, and make it redirect to a validation page on submit.

## Description

The add new email/phone button at the logins page now redirects to `/account/logins/[type]/new` and it displays `AddIdentity` component

It contains an input, an add button which should create the identity and redirect to `/account/logins/[type]/new/validation`, and a cancel button which should redirect to `/account/logins`.

By the way, I've decided to also make the `UpdateInput` redirect to `/account/logins/[type]/new/validation` on submit and the `DeleteButton` redirect to `/account/logins` on click.

I've implemented the `Validation` component, many of the buttons of this page are for now useless, and the `SnackBar` is appearing anytime without fadeout animation.

## How Has This Been Tested?

A new test file has been created with 2 new tests at `/tests/pages/newIdentity.test.tsx`, be careful, we now have a new warning in tests. Tests are still passing though

`(node:41083) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'push' of null
(node:41083) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:41083) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.` 

## Screenshots

image
image

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist:

- :green_circle: I have added tests to cover my changes.

[contributing_file]: https://github.com/fewlinesco/connect/blob/master/CONTRIBUTING.md


Fenntasy:
Task linked: [CU-7ev1pg Identity actions page (new)](https://app.clickup.com/t/7ev1pg)

src/components/UpdateInput.tsx
```diff
@@ -1,14 +1,17 @@
+import { useRouter } from "next/router";
```
Fuzznimp, line 1: You should remove the business logic from this component, and use the `business` and `display` folder.

So basically:

#### `/business/UpdateIdentity`
Which passes the async function used by the form component on submit as props to its children.

#### `/display/UpdateIdentityForm`
Which only display the css.

Fuzznimp: We could go even further and wrap `UpdateIdentityForm` with the add and remove business components. But it would require a bit of logic in the display component then (like the `preventDefault`).

I think the first one is better. @Fenntasy what do you think?

src/components/display/fewlines/SnackBar.tsx
```diff
@@ -0,0 +1,29 @@
...
+import styled from "styled-components";
+
+import { IdentityTypes } from "../../../@types/Identity";
+
+export const SnackBar: React.FC<{ type: IdentityTypes }> = ({ type }) => {
+  let method;
+  type.toUpperCase() === IdentityTypes.EMAIL
+    ? (method = "mail")
+    : (method = "SMS");
+  return <Snack>Confirmation {method} has been sent</Snack>;
```
Fuzznimp, line 11: Hmmm. I'll let fenn discuss this.

src/components/display/fewlines/SnackBar.tsx
```diff
@@ -0,0 +1,29 @@
...
+  margin-left: -125px;
+  background-color: #333;
+  color: #fff;
+  text-align: center;
+  border-radius: 2px;
+  padding: 16px;
+  position: fixed;
+  z-index: 1;
+  left: 50%;
+  bottom: 30px;
```
Fuzznimp, line 28: You need to use values from the theme. If you can't find values for you needs, feel free to add them. Just don't add random theme keys.

src/pages/account/logins/[type]/new.tsx
```diff
@@ -1,14 +1,62 @@
...
+    } catch (error) {
+      if (
+        error instanceof JsonWebTokenError ||
+        error instanceof NotBeforeError ||
+        error instanceof TokenExpiredError
+      ) {
+        Sentry.withScope((scope) => {
+          scope.setTag(
+            "/pages/account/logins SSR",
+            "/pages/account/logins SSR",
```
Comment (Fuzznimp) [inline src/pages/account/logins/[type]/new.tsx:47]: You need to change the scope tags to reflect where the error will happen.

src/utils/updateIdentity.ts
```diff
@@ -7,10 +7,10 @@ type UpdateBody = {
   value: string;
```
Fuzznimp, line 7: This should go inside the business component.

src/pages/account/logins/[type]/validation.tsx
```diff
@@ -0,0 +1,102 @@
...
+    } catch (error) {
+      if (
+        error instanceof JsonWebTokenError ||
+        error instanceof NotBeforeError ||
+        error instanceof TokenExpiredError
+      ) {
+        Sentry.withScope((scope) => {
+          scope.setTag(
+            "/pages/account/logins SSR",
+            "/pages/account/logins SSR",
```
Comment (Fuzznimp) [inline src/pages/account/logins/[type]/validation.tsx:66]: Change the tags.

src/pages/account/logins/[type]/validation.tsx
```diff
@@ -0,0 +1,102 @@
...
+import styled from "styled-components";
+
+import Sentry, { addRequestScopeToSentry } from "../../../..//utils/sentry";
+import { IdentityTypes } from "../../../../@types/Identity";
+import SnackBar from "../../../../components/display/fewlines/SnackBar";
+import { Button } from "../../../../pages/account/logins/index";
+
+const Validation: React.FC<{ type: IdentityTypes }> = (props) => {
+  const [validationCode, setValidationCode] = React.useState("");
+  const type = props.type;
```
Comment (Fuzznimp) [inline src/pages/account/logins/[type]/validation.tsx:21]:
```suggestion
const Validation: React.FC<{ type: IdentityTypes }> = ({type}) => {
  const [validationCode, setValidationCode] = React.useState("");
```

src/pages/account/logins/[type]/validation.tsx
```diff
@@ -0,0 +1,102 @@
+import { HttpStatus } from "@fewlines/fwl-web";
```
Comment (Fuzznimp) [inline src/pages/account/logins/[type]/validation.tsx:1]: You can move the form into `/display/IdentityValidationForm`. We won't do the business logic for now, it will be its own PR, with the validation workflow implementation.

src/pages/account/logins/[type]/new.tsx
```diff
@@ -1,14 +1,62 @@
...
 };
 
 export default NewIdentity;
 
-export const getServerSideProps: GetServerSideProps = async (_context) => {
-  return {
-    props: {},
-  };
-};
+export const getServerSideProps: GetServerSideProps = withSSRLogger(
```
Comment (Fuzznimp) [inline src/pages/account/logins/[type]/new.tsx:29]: Food for thought. When we'll do the middleware composition, we will need to abstract all the sentry/redirect part.

tests/pages/newIdentity.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+(useCookies as any).mockImplementation(() => {
+  return {
+    data: {
+      userId: "ac3f358d-d2c9-487e-8387-2e6866b853c9",
+    },
+  };
+});
+
+describe("the create form should work properly", () => {
+  test("it should display an input", () => {
```
Fuzznimp, line 41: Here is what I like to do, which is much clearer imo.

```suggestion
describe("AddNewIdentity", () => {
  test("should display an input", () => {
```

The top `describe` label just display the top component you are testing.

tests/pages/newIdentity.test.tsx
```diff
@@ -0,0 +1,79 @@
+jest.mock("../../src/hooks/useCookies");
```
Fuzznimp, line 1: The file naming should follow what you are testing. In this case, it should be `NewIdentity`.

But I think you should rename it too. This is a CRUD heavy application, so adding the actions in the components is crucial imo for scaling. So `AddNewIdentity` would be better I think.

Fuzznimp (changes requested):

src/components/UpdateInput.tsx
Fuzznimp: I just had the discussion with fenn, and we will use my second proposition, with a twist.

First, regarding the `preventDefault`, it is in fact the form responsibility to control its re-render. In other word, the business logic component do not care what kind of children they have, they only care about passing their logic. So the `preventDefault` would stay in the display component.

Like we said previously, the update identity flow is just a combination of `add` and `remove`. We already have those business component.

The update identity business component role would be to keep the current identity until it gets deleted.

So basically, `<UpdateIdentity/>` would be scoped inside the 3 business components.

src/components/display/fewlines/SnackBar.tsx
Fenntasy:
```suggestion
  const method = type.toUpperCase() === IdentityTypes.EMAIL ? "mail" : "SMS";
  return <Snack>Confirmation {method} has been sent</Snack>;
```
at least.

But I think we could create two versions:

```
  if (type.toUpperCase() === IdentityTypes.EMAIL) {
    return <Snack>Confirmation email has been sent</Snack>
  }
  if (type.toUpperCase() === IdentityTypes.PHONE) {
    return <Snack>Confirmation SMS has been sent</Snack>
  }
  return null
```

This might be more verbose but it's really easier to edit in the future and to replace by a translated string for each

Hamza-A: OK fixed! cd6f1c4

src/components/display/fewlines/SnackBar.tsx
Hamza-A: OK fixed! cd6f1c4

src/pages/account/logins/[type]/validation.tsx
Comment (Hamza-A) [reply src/pages/account/logins/[type]/validation.tsx]: OK fixed! ece8ff7

src/components/display/fewlines/SnackBar.tsx
Fuzznimp: Here is the one that still need to be used with theme values:

```suggestion
  visibility: visible;
  min-width: 250px;
  margin-left: Maybe this? Not sure.;
  background-color: ${({ theme }) => theme.colors.black};
  color: ${({ theme }) => theme.colors.background};
  text-align: center;
  border-radius: this;
  padding: this;
  position: fixed;
  z-index: 1;
  left: 50%;
  bottom: 30px;
`;
```

src/components/UpdateInput.tsx
```diff
@@ -1,14 +1,17 @@
...
 import React from "react";
 import styled from "styled-components";
 
 import { Identity } from "../@types/Identity";
 import { useCookies } from "../hooks/useCookies";
+import { Button } from "../pages/account/logins/index";
```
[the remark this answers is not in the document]
Hamza-A: OK for now I've just duplicated the styled button in this file because later we'll need to create several buttons with different styles (add, cancel, delete...) 8113d96

src/pages/account/logins/[type]/new.tsx
Comment (Hamza-A) [reply src/pages/account/logins/[type]/new.tsx:47]: OK fixed! a367648

src/pages/account/logins/[type]/validation.tsx
Comment (Hamza-A) [reply src/pages/account/logins/[type]/validation.tsx:65]: OK fixed! a367648

src/pages/account/logins/[type]/validation.tsx
Comment (Hamza-A) [reply src/pages/account/logins/[type]/validation.tsx:1]: OK fixed! 4a2f81f

tests/pages/newIdentity.test.tsx
Hamza-A: OK fixed d25506e

tests/pages/newIdentity.test.tsx
Hamza-A: OK fixed! d25506e

src/components/UpdateInput.tsx
Hamza-A: OK done! 0733f6a a8f1a4a e2787e7

src/components/display/fewlines/SnackBar.tsx
Hamza-A: I can't see `margin left` in theme options...

Fuzznimp: `margin` and `padding` are done with `theme.spaces`.

src/pages/account/logins/[type]/new.tsx
Comment (Fuzznimp) [reply src/pages/account/logins/[type]/new.tsx]: User id is sensitive data, we don't want to monitor them.

src/utils/updateIdentity.ts
Hamza-A: OK done 0733f6a

src/pages/account/logins/[type]/new.tsx
Comment (Fuzznimp) [reply src/pages/account/logins/[type]/new.tsx]: To extends my comment (as I was in the rush on the other repo).

It's ok to have the user ID in query strings, but we don't want to save it on our side (which will be the case if you add it in the trace, because we store them).

src/components/business/UpdateIdentity.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const UpdateIdentity: React.FC<UpdateIdentityProps> = ({
+  identity,
+  children,
+}) => {
+  const router = useRouter();
+
+  const { value } = identity;
+  let type: IdentityTypes = IdentityTypes.EMAIL;
+  identity.type === "phone" && (type = IdentityTypes.PHONE);
```
Fuzznimp, line 24:
```suggestion
  const { value, type } = identity;
  
  const castedIdentityType = type === ReceivedIdentityTypes.PHONE ? IdentityTypes.PHONE : IdentityTypes.EMAIL;
```

src/components/business/UpdateIdentity.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <AddIdentity type={type}>
+      {({ addIdentity }) => (
+        <DeleteIdentity type={type} value={value}>
+          {({ deleteIdentity }) =>
+            children({
+              updateIdentity: async function (newValue): Promise<void> {
+                addIdentity(newValue)
+                  .then(() => {
+                    deleteIdentity();
+                  })
```
Fuzznimp, line 36: We can do
```suggestion
                  .then(deleteIdentity)
```
right?

src/components/display/fewlines/AlertBar.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { IdentityTypes } from "../../../@types/Identity";
+
+export const AlertBar: React.FC<{ type: IdentityTypes }> = ({ type }) => {
+  if (type.toUpperCase() === IdentityTypes.EMAIL) {
+    return <Alert>Confirmation email has been sent</Alert>;
+  }
+  if (type.toUpperCase() === IdentityTypes.PHONE) {
+    return <Alert>Confirmation SMS has been sent</Alert>;
+  }
```
Fuzznimp, line 12:
```suggestion
  if (type === ReceivedIdentityTypes.EMAIL) {
    return <Alert>Confirmation email has been sent</Alert>;
  }
  if (type === ReceivedIdentityTypes.PHONE) {
    return <Alert>Confirmation SMS has been sent</Alert>;
  }
```

src/components/display/fewlines/IdentityValidationForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  &:active,
+  &:focus {
+    outline: none;
+  }
+`;
+
+const SendInput = styled.input`
+  color: ${({ theme }) => theme.colors.green};
+  padding: 0.25em 1em;
```
Fuzznimp, line 77: I would have say use the theme spacing, but we need to change the theme to better handle small spacing.

src/components/display/fewlines/SnackBar.tsx
Fuzznimp: Ok let's ignore that for now
https://github.com/fewlinesco/connect-account/pull/26/files#r483563388

src/pages/account/logins/[type]/new.tsx
Comment (Hamza-A) [reply src/pages/account/logins/[type]/new.tsx]: I only was using user ID in `src/pages/account/logins/[type]/[id]/[action].tsx`, but this file has been deleted afterwards 0ccaa8e

src/components/business/UpdateIdentity.tsx
Hamza-A: OK fixed! b9583db

src/components/business/UpdateIdentity.tsx
Hamza-A: Ok I did not know that before, thanks for the tip, fixed! b34faff

src/components/display/fewlines/AlertBar.tsx
Hamza-A: OK fixed! fb3d867 only the `.toUpperCase()` were useless

src/components/display/fewlines/IdentityValidationForm.tsx
Hamza-A: OK I ignore this for now

src/components/display/fewlines/SnackBar.tsx
Hamza-A: OK I ignore this for now

src/components/business/DeleteIdentity.tsx
```diff
@@ -32,8 +34,14 @@ export const DeleteIdentity: React.FC<DeleteIdentityProps> = ({
...
-  const deleteIdentity = (): Promise<Response> => {
-    return fetchJson("/api/account", HttpVerbs.DELETE, requestData);
+  const deleteIdentity = async (): Promise<void> => {
+    return fetchJson("/api/account", HttpVerbs.DELETE, requestData)
+      .then(() => {
+        router && router.push("/account/logins/");
+      })
+      .catch((error: Error) => {
+        console.warn(error.message);
+      });
```
Fuzznimp, line 44: You should throw an Error instead of a console.warn.

Hamza-A: OK fixed e5efe15

Fuzznimp (approved):

PR #27, opened 2020-09-08
Implement @fwl/oauth2

Fuzznimp:
This PR aims at using `@fwl/oauth2` to implement oauth2 flow and verify the jwts.

Here is an overview of what has been done:
- Adds the required env vars for the OAuth2Client initialization + config checks.
- Made use of oauth2client to verify acces tokens.
- Updated instance check of errors on SSR.
- `@fewlines/fwl-web` => `@fwl/web`.
- Fixed everything with the new router/pages.
- Partially add redirection if no cookie found.

On the partial redirection, I did not add it everywhere because it will be done with the middleware composition (auth middleware). The issue now is that, if you found yourself on a protected page, and don't have a cookie, this will break the app.

I have added the new config vars to Vercel.


.env_vars.sample
```diff
@@ -11,6 +11,10 @@ export CONNECT_PROVIDER_URL=***
 export CONNECT_APPLICATION_CLIENT_ID=***
 export CONNECT_APPLICATION_CLIENT_SECRET=***
 export CONNECT_APPLICATION_SCOPES=***
+export CONNECT_OPEN_ID_CONFIGURATION_URL=***
+export CONNECT_REDIRECT_URI=***
+export CONNECT_AUDIENCE=***
+export CONNECT_JWT_ALGO=***
```
Fenntasy, line 17:
```suggestion
export CONNECT_JWT_ALGORITHM=***
```

src/pages/account/index.tsx
```diff
@@ -22,6 +30,50 @@ const Account: React.FC = () => {
...
 
+export const getServerSideProps: GetServerSideProps = withSSRLogger(
+  withSession(async (context) => {
+    addRequestScopeToSentry(context.req);
+
+    try {
+      const accessToken = context.req.session.get("user-jwt");
+
+      if (accessToken) {
+        console.log(accessToken);
```
Fenntasy, line 41: 🔥

Fenntasy (approved):

PR #30, opened 2020-09-10
Identity verification + refacto add identity

Fuzznimp:
Todo: tests for the add identity workflow + verify identity.


lib/@types/Identity.ts
```diff
@@ -0,0 +1,26 @@
+export enum IdentityTypes {
```
Fenntasy, line 1: how could we make it so we don't need to update the lib each time someone on the backend team add an identity type on connect?

This is not a question for right now but I know they won't think of the lib without some kind of process and even with that, there could be misses. So we will need to keep that in mind (in my mind, that shouldn't be something that needs to be done but I don't know yet how we could do it)

lib/checkVerificationCode.ts
```diff
@@ -0,0 +1,32 @@
...
+  FetchResult<{ checkVerificationCode: CheckVerificationCodeResult }>
+> {
+  const operation = {
+    query: CHECK_VERIFICATION_CODE_QUERY,
+    variables: { code, eventId },
+  };
+
+  return (await fetchManagement(operation)) as FetchResult<{
+    checkVerificationCode: CheckVerificationCodeResult;
+  }>;
```
Fenntasy, line 31: This clear looks like a `return await` but I guess you can't do the `as FetchResult<...>` without it.
What do you think of storing the result in a variable that you type and then simply return this variable?

lib/checkVerificationCode.ts
```diff
@@ -0,0 +1,32 @@
+import { FetchResult } from "apollo-link";
```
Fenntasy, line 1: What about using CQRS inside the `lib` directory for the future? Separating queries and commands that way?

lib/sendIdentityValidationCode.ts
```diff
@@ -0,0 +1,48 @@
+import { FetchResult } from "apollo-link";
+import gql from "graphql-tag";
+
+import { fetchManagement } from "../src/utils/fetchManagement";
```
Fenntasy, line 4: Should we move this to `lib/`? or is it still used by something else in the app?

lib/checkVerificationCode.ts
Fenntasy: better yet, what about making?
```typescript
fetchManagement<{
     checkVerificationCode: CheckVerificationCodeResult;
   }>(operation)
```

src/@types/mongo/User.ts
```diff
@@ -10,6 +10,7 @@ export type TemporaryIdentity = {
   eventId: string;
   value: string;
   type: string;
+  ttl: number;
```
Fenntasy, line 13: TTL is supposed to be a duration, how do you know when it has expired?

Wouldn't it be simpler with `expiresAt`?

src/errors.ts
```diff
@@ -17,3 +18,16 @@ export function OAuth2Error(error?: OAuth2Errors): WebError {
...
+export class MongoNoDataReturned extends Error {}
+
+export class GraphqlErrors extends Error {
+  parentErrors: string;
+
+  constructor(errors: readonly GraphQLError[]) {
+    super();
+    this.parentErrors = errors.map(({ message }) => message).join(" | ");
+  }
+}
```
Fenntasy, line 31: Here's a proposal to keep the idea of having only `parentError`

```suggestion
export class GraphqlError extends Error {
  parentError: GraphqlErrors;
  
  constructor(message, error: readonly GraphqlErrors) {
    super(message)
    this.parentError = error;
  }
}

export class GraphqlErrors extends Error {
  constructor(errors: readonly GraphQLError[]) {
    const message = errors.map(({ message }) => message).join("\n")
    super(message);
  }
}
```

and then, instead of 

```typescript
  if (result.errors) {
    throw result.errors;
  }
```

```typescript
  if (result.errors) {
    throw new GraphqlErrors(result.errors);
  }
```

That could be catched up the chain by a `new GraphqlError("graphql error", error)`

lib/@types/Identity.ts
Fuzznimp: I keep thinking about it. I think it's a quite complex task to do. We would need to generate types/enums from the GraphQL schemas (should be doable). The hard part would be to keep track of the schemas update, to re-generate the types/enums, and publish after that. I have no clear solution for that, except for a GitHub template with a checkbox regarding identities type update, and if checked, this should be reported to the team maintaining the lib. Looks like a pretty heavy and error prone solution.

lib/checkVerificationCode.ts
Fuzznimp: I wanted to make `fetchManagenement()` typed with a generic a few weeks ago, but it looks like a complex task:

```ts
Type 'Promise<FetchResult<{ [key: string]: any; }, Record<string, any>, Record<string, any>>>' is not assignable to type 'Promise<FetchResult<T, Record<string, any>, Record<string, any>>>'.
  Type 'FetchResult<{ [key: string]: any; }, Record<string, any>, Record<string, any>>' is not assignable to type 'FetchResult<T, Record<string, any>, Record<string, any>>'.
    Type 'FetchResult<{ [key: string]: any; }, Record<string, any>, Record<string, any>>' is not assignable to type 'ExecutionResult<T>'.
      Types of property 'data' are incompatible.
        Type '{ [key: string]: any; } | null | undefined' is not assignable to type 'T | null | undefined'.
          Type '{ [key: string]: any; }' is not assignable to type 'T'.
            'T' could be instantiated with an arbitrary type which could be unrelated to '{ [key: string]: any; }'.ts(2322)
```

Fenntasy: oook 😅

lib/checkVerificationCode.ts
Fuzznimp: I agree, but I'm not sure about where to put what thought in the case of GraphQL. Am I right to think that queries goes into queries, and mutations/subscriptions are commands?

lib/sendIdentityValidationCode.ts
Fuzznimp: `fetchManagement` is used in pages' getServerSideProps to retrieve identities.

Fenntasy: ok, so in a next PR when those move then 😁

Fuzznimp: Do you think the lib should handle getting the identities?

src/@types/mongo/User.ts
Fuzznimp: `expiresAt` should make the code more readable.

lib/sendIdentityValidationCode.ts
Fenntasy: Yes, the lib should handle all relation with Connect

Fenntasy: and Connect Profile even

src/errors.ts
Fuzznimp: I like it!

Fuzznimp:
@Fenntasy I realized while writing the documentation that I forgot to handle the remove of temporary identities from the DB. As we are still not sure if we will stick to mongo, do you want me to do it in here, or should I just make a task to handle it?

Fuzznimp:
@Fenntasy Well, I think it's a high priority to merge this PR, as it is preventing Hamza from going further (because of the massive conflicts it brings). I'm making a task.

src/components/display/fewlines/AddIdentityInputForm.tsx
```diff
@@ -24,13 +29,20 @@ export const AddIdentityInputForm: React.FC<AddIdentityInputFormProps> = ({
...
+
+          const eventId = await fetchJson(
+            "/api/auth-connect/send-identity-validation-code",
+            HttpVerbs.POST,
+            body,
+          ).then((data) => data.json());
+
+          router &&
+            router.push(`/account/logins/${type}/validation/${eventId.data}`);
         }}
```
Fenntasy, line 46: Should that be in a display component?

src/components/display/fewlines/IdentityValidationForm.tsx
```diff
@@ -2,26 +2,44 @@ import { useRouter } from "next/router";
...
+          event.preventDefault();
+
+          await fetchJson(
+            "/api/auth-connect/verify-validation-code",
+            HttpVerbs.POST,
+            { validationCode, eventId },
+          );
+
+          router && router.push("/account/logins");
+        }}
```
Fenntasy, line 35: Same question: in a display component?

src/pages/api/auth-connect/send-identity-validation-code.ts
```diff
@@ -0,0 +1,93 @@
...
+      const user = await getUser(request.headers["cookie"] as string);
+
+      if (user) {
+        const decoded = await oauth2Client.verifyJWT<{ sub: string }>(
+          user.accessToken,
+          config.connectJwtAlgorithm,
+        );
+
+        const identity = {
+          status: IdentityStatus.VALIDATED,
```
Fenntasy, line 32: Why is it validated before sending the code? 🤔

src/pages/api/auth-connect/verify-validation-code.ts
```diff
@@ -0,0 +1,75 @@
...
+            const absoluteURL = new URL(route, config.connectDomain).toString();
+
+            await fetchJson(absoluteURL, HttpVerbs.POST, body);
+
+            response.writeHead(HttpStatus.TEMPORARY_REDIRECT, {
+              Location: "/account/logins",
+            });
+
+            response.end();
+          }
```
Fenntasy, line 52: else?

tests/pages/AddIdentityInputForm.test.tsx
```diff
@@ -52,7 +52,9 @@ describe("AddNewIdentity", () => {
     expect(addIdentityInput).toHaveLength(1);
   });
 
-  test("it should submit the form", () => {
+  test("it should call `send-identity-validation-code` API page on submit", () => {
+    Date.now = jest.fn(() => 1572393600000);
```
Fenntasy, line 56:
```suggestion
    const spyDate = jest.spyOn(Date, "now").mockReturnValue(1572393600000);
```

with:

```
spyDate.mockRestore()
```

at the ned of the test

tests/pages/IdentityValidationForm.test.tsx
```diff
@@ -110,4 +132,40 @@ describe("IdentityValidationForm", () => {
...
+
+    const validationCode = "42";
+
+    verifyIdentityInput.simulate("change", {
+      target: { value: validationCode },
+    });
+
+    const form = component.find(IdentityValidationForm).find(Form);
+
+    const fetchJsonMethod = jest.spyOn(fetchJson, "fetchJson");
```
Fenntasy, line 158: I wouldn't have thought about this method but 👍

src/@types/Identity.ts
```diff
@@ -1,16 +1,3 @@
-export enum IdentityTypes {
```
Fenntasy, line 1: We don't need them anymore?

src/components/display/fewlines/AddIdentityInputForm.tsx
Fuzznimp: I put that here because I thought previously that we were getting rid of the `RenderProps` usage.
I will extract that into its business component 👍.

Fuzznimp: Now that I think about, I should make a task and do that somewhere else, as I also need to refacto the `AddNewIdentity` business component. What do you think?

Fenntasy: As long as it doesn't stay too long, I'm ok with it 😁

src/pages/api/auth-connect/send-identity-validation-code.ts
Fuzznimp: I hard coded the validation because, if the identity isn't verified with the verification flow, the add identity mutation isn't trigger. Let me know if it does make sense.

src/pages/api/auth-connect/verify-validation-code.ts
Fuzznimp: Are you ok with just `throw new WrongVerificationCodeInput();` (i.e. not adding a custom error message).

Fenntasy: Well, for a start, it should reask for the verification code in a loop, so I would imagine a redirect

Fuzznimp: You are right, I totally forgot about that..

Fuzznimp: I'll need to implement the `flash` logic soon imo. Hamza also needs it for the `AlertBar` triggers.

Fenntasy (approved):

PR #46, opened 2020-09-29
Refacto input logic into business component

Fuzznimp:
## Description

This PR aims at refactoring the inputs logic into its own component.

We have two implementation of the `AddIdentity` (the new one and the old one renamed `AddIdentity_Deprecated`). This PR only aims at the "add" flow, and not the "update" flow, hence the two instances.

## How Has This Been Tested?

Manually and with Jest

## Types of changes

- Chore (non-breaking change which refactors / improves the existing code base).
- ~Bug fix (non-breaking change which fixes an issue).~
- ~New feature (non-breaking change which adds functionality).~
- ~Breaking change (fix or feature that would cause existing functionality to
  change).~

## Checklist:

- ✅ My code follows the code style of this project.
- 🔴 My change requires a change to the documentation.
- 🔴 I have updated the documentation accordingly.
- ✅ I have read the [**CONTRIBUTING**][contributing_file] document.
- 🔴 I have added tests to cover my changes.
- ✅ All new and existing tests passed.

### DIFF
--- a/src/components/business/AddIdentity.tsx
+++ b/src/components/business/AddIdentity.tsx
@@ -1,41 +1,41 @@
+import { useRouter } from "next/router";
 import React from "react";
 
 import { HttpVerbs } from "@src/@types/HttpVerbs";
-import type { ReceivedIdentityTypes } from "@src/@types/Identity";
-import { useCookies } from "@src/hooks/useCookies";
+import type {
+  InMemoryTemporaryIdentity,
+  ReceivedIdentityTypes,
+} from "@src/@types/Identity";
 import { fetchJson } from "@src/utils/fetchJson";
 
 interface AddIdentityProps {
   type: ReceivedIdentityTypes;
   children: (props: {
-    addIdentity: (value: string) => Promise<Response>;
-    type: ReceivedIdentityTypes;
+    addIdentity: (identity: InMemoryTemporaryIdentity) => Promise<void>;
   }) => JSX.Element;
 }
 
 export const AddIdentity: React.FC<AddIdentityProps> = ({ type, children }) => {
-  const { data, error } = useCookies();
+  const router = useRouter();
 
-  if (error) {
-    return <div>Failed to load</div>;
-  }
-
-  if (!data) {
-    return null;
-  }
-
-  const addIdentity = (value: string): Promise<Response> => {
+  async function addIdentity(
+    identity: InMemoryTemporaryIdentity,
+  ): Promise<void> {
     const body = {
-      userId: data.userSub,
-      type: type.toUpperCase(),
-      value,
+      callbackUrl: "/",
+      identityInput: identity,
     };
 
-    return fetchJson("/api/identities", HttpVerbs.POST, body);
-  };
+    const eventId = await fetchJson(
+      "/api/auth-connect/send-identity-validation-code",
+      HttpVerbs.POST,
+      body,
+    ).then((data) => data.json());
+
+    router && router.push(`/account/logins/${type}/validation/${eventId.data}`);
+  }
 
   return children({
     addIdentity,
-    type,
   });
 };
--- a/src/components/business/AddIdentity.tsx
+++ b/src/components/business/AddIdentity.tsx
@@ -6,6 +6,7 @@ import type {
   InMemoryTemporaryIdentity,
   ReceivedIdentityTypes,
 } from "@src/@types/Identity";
+import { ErrorSendingValidationCode } from "@src/clientErrors";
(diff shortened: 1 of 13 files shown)


src/components/business/AddIdentity.tsx
```diff
@@ -1,41 +1,41 @@
...
 
-    return fetchJson("/api/identities", HttpVerbs.POST, body);
-  };
+    const eventId = await fetchJson(
+      "/api/auth-connect/send-identity-validation-code",
+      HttpVerbs.POST,
+      body,
+    ).then((data) => data.json());
+
+    router && router.push(`/account/logins/${type}/validation/${eventId.data}`);
```
alli5723, line 35: While reading through, I was wondering what would happen on 2 occasions where `fetchJson` does not return an expected response (e.g. returning a null or an empty object or some other unexpected response). 
What will happen to `router.push` in this case? 🤔

Fuzznimp: That's a good question. Passing anything but a path to the router will break the app in dev environment, or will call a fallback component in production.

Just so you know - and you probably have seen it - the error handling is a big weak point in the application at the moment. I'm still not sure when I should start implementing a solid error handling system (with error stack).

To go back on the topic, this is the kind of situation where I lack the experience on what behaviour should be triggered in this case. Should it break the app? Should it check for the truthly nature of the response, and default redirect to, in this case, to the same page (i.e. reload the page).

Let me know what you think.

alli5723: After giving some thoughts to this, I think reloading the same page with some message for the user to retry might be good, but this would seem fine if the error situation is only temporary.
I guess that is where the error handling you mentioned comes in, so we can at least classify the type of error received and handle redirection appropriately based on the error type.

For now, if we could simply check for Error 5XX and 4XX to handle this redirection do you think it will suffice?

Fuzznimp: Yes, I will definitely add a status code check in the handler.

About the message though, the flash feature will come later, when I'll have time to do it 👍.

Fuzznimp: 8e7a793 Would it be enough for you?

alli5723: 👍  ... Looks goods.

I am not sure ` return router && router.push("/account/logins/email/new");` in the file `src/components/business/AddIdentity.tsx` will terminate the execution. Are we sure the execution will not continue after redirecting user to `/account/logins/email/new`? I was thinking of an alternative as written below.

Sorry, if I have made any mistake in the suggestion 🤦 
```
await fetchJson(
  "/api/auth-connect/send-identity-validation-code",
  HttpVerbs.POST,
  body,
).then((response) => {
  if (response.status >= 400) {
    return null;
  }

  return response.json();
}).then((eventIdData) => {
  if (eventIdData) {
    router && router.push(`/account/logins/${type}/validation/${eventIdData.data}`);
  } else {
    router && router.push("/account/logins/email/new");
  }
});
```

Fuzznimp: > I am not sure return router && router.push("/account/logins/email/new"); in the file src/components/business/AddIdentity.tsx will terminate the execution.

I have a doubt now haha. @Fenntasy do you think the function will still return `response.json()` if `response.status >= 400`?

```ts
    ).then((response) => {
      if (response.status >= 400) {
        return router && router.push("/account/logins/email/new");
      }

      return response.json();
    });
```

Fuzznimp: So we had the discussion with fenn.

In the snippet above, if `response.status >= 400` was true, it would not return `response.json()`.

A better way to handle it is to throw an error if `response.status >= 400` is `true`, catch it and redirect.

8209ba6

I had to put the custom error `ErrorSendingValidationCode` in a new file, because importing and using an error from `src/errors` on the client side would throw:

```
> Build error occurred
Error: > Build failed because of webpack errors
    at build (/Users/fuzznimp/workspace/fewlines/connect-account/node_modules/next/dist/build/index.js:13:917)
```

alli5723: Yea, I am actually happy with throwing the error 🙂

Fuzznimp: I've also added this response handling t the verify validation code flow: 1d56ac1.
This PR can be re-reviewed.
There is an issue with the CI checks at the bottom (i.e. `Vercel Expected — Waiting for status to be reported`). This probably have to do with the CI building and deploying storybook. I'll ping you when I'm done investigating.

alli5723 (approved):

PR #49, opened 2020-10-01
using getIdentity instead of getIdentities

Hamza-A:
This branch aims to implement a new GraphQL query `getIdentity` to find a single identity with an `id` on params instead of using `getIdentity` and then filter on results.

## Description
I have created a new file `@lib/query/getIdentity` in which the query has been stored, this query is now used in the `getServerSideProps` of `src/pages/account/logins/[type]/[id]` page. I have changed `type ProviderUser` accordingly.

## How Has This Been Tested?
There were no need to add or modify tests

## Types of changes

- Chore (non-breaking change which refactors / improves the existing code base).

## Checklist:

- :red_circle: I have added tests to cover my changes.


lib/queries/getIdentity.ts
```diff
@@ -0,0 +1,38 @@
+import { FetchResult } from "apollo-link";
+import gql from "graphql-tag";
+
+import type { ProviderUser } from "@src/@types/ProviderUser";
+import { fetchManagement } from "@src/utils/fetchManagement";
+
+const USER_QUERY = gql`
```
Fuzznimp, line 7: On this subject, the query document in `getIdentities` should have been called `GET_USER_IDENTITIES_QUERY`, so this one should be

```suggestion
const GET_USER_IDENTITY_QUERY = gql`
```

Could you also change the name in `getIdentities`? Thanks!

src/@types/ProviderUser.ts
```diff
@@ -6,5 +6,6 @@ export type ProviderUser = {
   user: {
     id: string;
     identities: Identity[];
+    identity: Identity;
   };
 };
```
Fuzznimp, line 11: Hmm.. I think this should be unchanged. The query should still return `Identity[]`, but with only one identity in it, based on the filter. Could you verify this?

Fuzznimp (changes requested):

lib/queries/getIdentity.ts
Hamza-A: OK fixed 7714de7

src/@types/ProviderUser.ts
Hamza-A: It returns an object
`{
  data: {
    provider: {
      id: 'xxxx-xxxx-xxxx-xxxx',
      name: 'Bluesports',
      user: [Object]
    }
  }
}`

Fuzznimp: It's the `user` field that is useful to show :).

Hamza-A: It's still an object
`{
  id: 'xxxx-xxxx-xxxx-xxxx',
  identity: {
    id: 'xxxx-xxxx-xxxx-xxxx',
    primary: true,
    status: 'validated',
    type: 'email',
    value: 'xxx.xxx@xxx.xxx'
  }
}`

Fuzznimp: You can go deeper in the object by calling it in your console.log.

lib/queries/getIdentity.ts
Hamza-A: OK fixed 69a9c93

src/@types/ProviderUser.ts
Hamza-A: OK fixed 69a9c93

Fuzznimp (approved):

PR #53, opened 2020-10-06
Cu 8cnc97 css integration confirmation box

Hamza-A:
This PR aims to implement a new confirmation box which appears whenever a user tries to delete an identity or modify his primary one

## Description
To do this, I had created 2 new components:
- the `<ConfirmationBox />` itself
- a `<ClickAwayListener />` which should make the `<ConfirmationBox />` disappear on click

I have created stories with these 2 components gathered.

These components has been implemented in `<ShowIdentity />` twice:
- for the make primary button
- for the delete button
And I had to create a new state to control the visibility of the `<ConfirmationBox />`

## How Has This Been Tested?
I added 8 new tests in `ShowIdentityPage.test.tsx` in a new `describe()` called `ConfirmationBox`

## Screenshots
image
image

## Types of changes

- New feature (non-breaking change which adds functionality).

## Checklist:

- :green_circle: I have added tests to cover my changes.


src/components/display/fewlines/ConfirmationBox/ConfirmationBox.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const ConfirmationBox = styled.div<ConfirmationBoxProps>`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 4rem 2rem 3rem;
+  position: fixed;
+  background-color: ${({ theme }) => theme.colors.background};
+  width: 100%;
+  left: 0;
+  ${(props) =>
```
Fuzznimp, line 16:
```suggestion
  position: fixed;
  left: 0;
  background-color: ${({ theme }) => theme.colors.background};
  width: 100%;

  ${(props) =>
```

It is a good practice to put the positioning next to the `position: fixed`.

src/components/display/fewlines/ShowIdentity/ShowIdentity.tsx
```diff
@@ -60,18 +70,72 @@ export const ShowIdentity: React.FC<ShowIdentityProps> = ({ identity }) => {
...
+          </Button>
+          {primaryConfirmationBoxOpen && (
+            <ClickAwayListener
+              onClick={() => setPrimaryConfirmationBoxOpen(false)}
+            />
+          )}
+          <ConfirmationBox open={primaryConfirmationBoxOpen}>
+            <p className="confirmation-text">
+              You are about to replace mail@mail.com as your main address
+            </p>
```
Fuzznimp, line 88: Maybe make this as a `<ConfirmationText/>`. What do you think?

src/components/display/fewlines/ConfirmationBox/ClickAwayListener.tsx
```diff
@@ -0,0 +1,10 @@
+import styled from "styled-components";
+
+export const ClickAwayListener = styled.div`
+  position: absolute;
+  height: 100%;
+  z-index: 1;
+  width: 100%;
+  left: 0;
+  top: 0;
+`;
```
Fuzznimp, line 10: @Fenntasy What do you think of this (i.e. a `styled-component` with a `onClick` call)? 

Would you be ok with custom hook like this?

```ts
import React from 'react';

/**
 * Run the specified function whenever the user clicks somewhere outside of the
 * surface area of the supplied reference.
 *
 * Example: Closing a modal when the user clicks away from it.
 * @param ref RefObject of the element we want to react to outside clicks for.
 * @param handler The function to be executed when an outside click is registered.
 */
const useOnClickOutside = (
	ref: React.RefObject<any>,
	handler: (e: React.MouseEvent) => void
): void => {
	React.useEffect(() => {
		const listener = (event: React.MouseEvent) => {
			if (!ref.current || ref.current.contains(event.target)) {
				return;
			}

			handler(event);
		};

		document.addEventListener('mousedown', listener as any);

		return () => {
			document.removeEventListener('mousedown', listener as any);
		};
	}, [ref, handler]);
};

export default useOnClickOutside;
```

tests/pages/ShowIdentityPage.test.tsx
```diff
@@ -185,32 +167,26 @@ describe("ShowIdentityPage", () => {
...
     );
 
     const primaryBadge = component.contains(<PrimaryBadge />);
-    const makeThisPrimaryButton = component.contains(
-      <Button variant={ButtonVariant.SECONDARY}>
-        Make this my primary email
-      </Button>,
-    );
+    const makeThisPrimaryButton = component
+      .find(Button)
```
Fuzznimp, line 177: I guess that `component.find()` returns **all** the `Button` it finds, right?

src/components/display/fewlines/ConfirmationBox/ConfirmationBox.tsx
Hamza-A: OK fixed 868fb95

src/components/display/fewlines/ShowIdentity/ShowIdentity.tsx
Hamza-A: OK I created a new component `<ConfirmationText />` in its own file and exported to stories, `<ShowIdentity />` component and tests 693fcec

src/components/display/fewlines/ConfirmationBox/ClickAwayListener.tsx
Fenntasy: I'm not against the principle but I'm always afraid of this kind of stuff: 
- do we support enough browser with it? 
- what happens if we click outside a modal but on a link, do we follow the link because the event has bubbled down?

There's always some edge cases that are annoying (with this particular stuff) that are a pain to debug.
That's why I always rely on a dom element that I create behind my modal, because I have more control over it.

But I insist, it's this particular case that makes me affraid, not this kind of custom hook

Fuzznimp: @Hamza-A So after some discussion, we decided to keep your version.

Also
```suggestion
import styled from "styled-components";

export const ClickAwayListener = styled.div`
  position: fixed;
  width: 100%;
  height: 100vh;
  z-index: 1;
`;
```

Hamza-A: OK I kept `left: 0;` and `top: 0;` though, Otherwise, the listener would not cover the whole screen as illustrated by the red part 
image
a5a85c9

Fuzznimp: > OK I kept left: 0; and top: 0; though

Ah yes, my bad.

tests/pages/ShowIdentityPage.test.tsx
Hamza-A: Yes, I had to implement this way to find the button since it now has an `onClick` property that would be tricky to mock in this test file.
However, I'm checking on line 182 that the button I'm testing is the one I'm expecting

src/components/display/fewlines/ShowIdentity/ShowIdentity.tsx
Fuzznimp: Do you think you will re-use this component? If so then you can leave it in its own file.

Hamza-A: I don't think I'll use it outside of a `<ConfirmationBox />`, may I move those 2 components to the same file?

Fuzznimp: Yeah, you can put that back in `ShowIdentity`.

Hamza-A: I would stand for putting it in the same file as `<ConfirmationBox />` instead. Since I'm using `<ConfirmationText />` also in the stories.
Won't it look weird if I import a component from a page ? What is your opinion?

Fuzznimp: It also feels weird declare something and not use it in the current file, don't you think? This would just create an extra import in `ShowIdentity`.

Hamza-A: OK fixed d53c644

Fuzznimp (approved):

PR #61, opened 2020-10-12
Cu 92wyte setup taiko

Fuzznimp:
## Description

This PR aims at discovering and validating using `taiko` as your e2e test tool.

## Type of change

## Issues

So far, using `Puppeteer` is a real pain, both in terms of setting constraints and test writing. `taiko` is probably the way to go.

## How Has This Been Tested?

Automated tests

## List of added dependencies

- "taiko": "1.0.24",

## Checklist:


tests/e2e/show-identity.test.ts
```diff
@@ -0,0 +1,124 @@
...
+      expect(text("Login").exists()).toBeTruthy();
+    } catch (error) {
+      await screenshot({ path: "tests/e2e/screenshots/launch.png" });
+
+      throw error;
+    }
+  });
+
+  describe("Perform the authentification", () => {
+    test("click the login button", async () => {
```
Fenntasy, line 40: It's really weird that it works like that. Tests should be able to be ran in any order and having them go sequentially is more of a bug than a feature. 
I realize it allows for simpler `try/catch` blocks but I think we should aim for another method, what do you think?

Fuzznimp: Well, I did it to have a better time seeing where the issue is in case of an exception thrown. I don't really get why this would be a bug though.
I can do it in one `test` block. I have some ideas on how to improve the the writing of tests (through utils).
This was more of a POC. For info, this PR took me what.. 15min? Sooooo.. can we use it instead of puppeteer 🙇?

Fenntasy: > Well, I did it to have a better time seeing where the issue is in case of an exception thrown. I don't really get why this would be a bug though.

Yup, I get why you did ^^. The bug is that tests in a test file could be launched at the same time and the order is not guaranteed so... we could have them run either at the same time or in a wrong order and it could be a bit weird to debug.

Fenntasy: > This was more of a POC. For info, this PR took me what.. 15min? Sooooo.. can we use it instead of puppeteer 🙇?

Well what do you think? I haven't tested it so I rely on your judgment

Fuzznimp: > The bug is that tests in a test file could be launched at the same time and the order is not guaranteed so

Now that I think about it, `--runInBand` would enforce tests being run sequentially. What do you think?

Fenntasy: Technically, it enforce that they won't be run in parallel, they don't enforce the order

Fenntasy (approved):

PR #129, opened 2020-11-23
Add/remove redirect_uris to Management in GH actions

Oluane:
## Description

The goal of this PR is to implement first part of e2e tests launch in GH actions. This 1st part is about getting Vercel preview deployment's url, adding it to the test Application's redirect uris in Management and removing it. 

## Type of change

## List of added dependencies

Added `core-fetch` dependency, to have a fetch API in Node env

## Checklist:


bin/addRedirectURItoConnect.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const githubActionsContext = JSON.parse(process.env.GITHUB_CONTEXT_EVENT);
+
+    if (githubActionsContext.deployment_status === undefined)
+      throw new Error("deploymentStatus is undefined");
+
+    const vercelDeployment = githubActionsContext.deployment_status;
+
+    if (vercelDeployment.state === "success") {
+      const testApp = await getApplication(config.connectAccountTestAppId).then(
```
Fuzznimp, line 18: I like what you did with your truthly checks on the env vars below. Now that I think about it, I was probably wrong when I told you to add this to the config object (because it only concerns GitHubs Actions env). What do you think?

bin/addRedirectURItoConnect.ts
```diff
@@ -0,0 +1,54 @@
+import { updateApplication } from "../lib/commands/updateApplication";
+import { getApplication } from "../lib/queries/getApplication";
+import { config } from "../src/config";
+
+const addingRedirectURIToConnect = async (): Promise<any> => {
```
Fuzznimp, line 5:
```suggestion
const addingRedirectURIToConnect = async (): Promise<void> => {
```
Same on the other one.

bin/addRedirectURItoConnect.ts
```diff
@@ -0,0 +1,54 @@
...
+      await updateApplication(updatedTestApp).then((results) => {
+        if (results.errors) {
+          throw results.errors;
+        }
+
+        if (!results.data) {
+          throw new Error("update error");
+        }
+
+        return results.data;
```
Fuzznimp, line 46: Are you sure you need to return this :)?

lib/@types/Application.ts
```diff
@@ -0,0 +1,19 @@
...
+  redirectUris: string[];
+};
+
+export type ApplicationInput = {
+  id: string;
+  description: string;
+  name: string;
+  redirectUris: string[];
+  defaultHomePage: string;
+};
```
Fuzznimp, line 15: I understand why you did this, but we probably can ride on one of the two :). Not sure what to think though.

lib/commands/updateApplication.ts
```diff
@@ -0,0 +1,41 @@
+import { FetchResult } from "apollo-link";
+import gql from "graphql-tag";
+
+import { fetchManagement } from "../../src/utils/fetchManagement";
+import { ApplicationInput } from "@lib/@types/Application";
```
Fuzznimp, line 5: Wait, is this path alias working during your script execution?

lib/queries/getApplication.ts
```diff
@@ -0,0 +1,32 @@
...
+import gql from "graphql-tag";
+
+import { fetchManagement } from "../../src/utils/fetchManagement";
+import type { ProviderApplication } from "@lib/@types/Application";
+
+const GET_APPLICATION_QUERY = gql`
+  query getApplicationQuery($id: String!) {
+    provider {
+      application(filters: { id: $id }) {
+        id
```
Fuzznimp, line 11: Do you think it's a good idea to return the id?

lib/commands/updateApplication.ts
```diff
@@ -0,0 +1,41 @@
...
+      redirectUris
+      name
+      defaultHomePage
+    }
+  }
+`;
+
+export async function updateApplication(
+  command: ApplicationInput,
+): Promise<FetchResult> {
```
Fuzznimp, line 34: You should type the return value of `FetchResult`.

package.json
```diff
@@ -15,7 +15,9 @@
     "build-storybook": "build-storybook",
     "db:create-users-table": "ts-node ./src/scripts/create_users_table.ts",
     "db:delete-users-table": "ts-node ./src/scripts/delete_users_table.ts",
-    "db:get-user-from-sub": "ts-node ./src/scripts/get_user_from_sub.ts"
+    "db:get-user-from-sub": "ts-node ./src/scripts/get_user_from_sub.ts",
+    "e2e-addRedirectUri": "ts-node bin/addRedirectURItoConnect.ts",
+    "e2e-removeRedirectUri": "ts-node bin/removeRedirectURIFromConnect.ts"
```
Fuzznimp, line 20: Do you think we should keep consistency in our script naming? Something like:
```suggestion
    "e2e:add-redirect-uri": "ts-node bin/add-redirect-uri-from-application.ts",
    "e2e:remove-redirect-uri": "ts-node bin/remove-redirect-uri-from-application.ts"
```
That way, if we see kebab casing, we know it's a "self invoked" function.

Fuzznimp (changes requested):

yarn.lock
```diff
@@ -6972,7 +6972,7 @@ cross-fetch@3.0.5:
   dependencies:
     node-fetch "2.6.0"
 
-cross-fetch@^3.0.4:
+cross-fetch@3.0.6, cross-fetch@^3.0.4:
```
Fenntasy, line 6975: why not `node-fetch`?

Oluane: We used that one on the first place, but there was a compatibility issue with `apollo-link-http`.

Fenntasy: oh, what was the problem?

Oluane: In `fetchMangement`, we construct a `httpLink` with `apollo-link-http` for GraphQL queries & mutations, with API url & `fetch`. But as the new scripts launched in GH actions are running in Node env, we can't use the global `fetch`provided by NextJs, so we needed an external one. And sadly, `node-fetch` was not compatible with `apollo-link-http` (more precisely, it seems to be a type issue).

package.json
Oluane: I agree, let's do that ! It will be more clear that way.

bin/addRedirectURItoConnect.ts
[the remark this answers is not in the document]
Oluane: Yes ahah

bin/addRedirectURItoConnect.ts
Oluane: Nope, indeed.

yarn.lock
Fuzznimp: https://github.com/apollographql/apollo-link/issues/513#issuecomment-384743835

Fuzznimp: We could add `isomorphic-fetch` too if you prefer Fenn.

bin/addRedirectURItoConnect.ts
Oluane: Yep, since this env var is only used in GH actions, I think there's no need to add it on project's config object (and it may be a bit confusing to see it in config obj, but used nowhere in the projects files). But I may be wrong. :)

Fuzznimp: You are absolutely right :)

bin/addRedirectURItoConnect.ts
```diff
@@ -0,0 +1,54 @@
...
+      });
+    }
+  } catch (error) {
+    console.log(error);
```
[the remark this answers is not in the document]
Oluane: Now that you say that, I wonder if the try-catch thing is really needed there ? What do you think ?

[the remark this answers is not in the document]
Fuzznimp: Nop, there is no need for it if we don't do anything with the error, except throwing it (which you already do in the `try` scope :)).

lib/queries/getApplication.ts
Oluane: I did it that way because I use the `getApplication` response to update the application, and `updateApplication` needs the Application id. But, I think I could add that id myself to the App object.

Fuzznimp: I'm worried that this sensible data could be catch in clear if we do have it in the response. Am I wrong @Fenntasy?

bin/addRedirectURItoConnect.ts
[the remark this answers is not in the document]
Oluane: Alright :)

lib/commands/updateApplication.ts
Oluane: That's strange, but this one works. But for consistency, I may use relative path I think.

yarn.lock
Fenntasy: Well no, let's go with `cross-fetch` since it's just an `if` around `node-fetch` anyway. I'm annoyed at the appolo team since https://github.com/apollographql/apollo-link/issues/513#issuecomment-436364057 
I don't know where the faulty types lies (could be from apollo, from node-fetch or even be the wrong type in both) but the fact that they rely on the fact that `cross-fetch` or `isomorphic-fetch` have a a type that goes better with their code and don't work with the other ones is not a good look in my opinion.

That being said, if `cross-fetch` does the job ¯\_(ツ)_/¯

lib/queries/getApplication.ts
Fenntasy: hmmm I don't really see the ID as a sensitive data, especially if we only use this query in a script that run on the CI.

But what strikes me as weird is to have this code in the `lib` part of Account when it's (I suppose?) only used in a script. I don't think we should, in the application do those kind of request, would we?

yarn.lock
Fuzznimp: I agree, good thing we are moving away from GraphQL for Account then I guess.

lib/queries/getApplication.ts
Fuzznimp: Where do you think we should put these scripts then?

Couple questions to clarify. Do we want to permit Clients (from clients, not us) to create/delete applications? If we do, why couldn't it be in the NodeSDK?
And if we don't, where should we put this?

Fenntasy: that's a good point, my mistake 😅

lib/commands/updateApplication.ts
Oluane: Yes sir ahah

Fuzznimp (approved):

PR #154, opened 2020-12-08
Fix IdentityType error

Oluane:
## Description

This PR aims to fix the fact that an exception was thrown when an unknown `IdentityType` was provided by `Management` to `Account` app and broke the app. 

Now, we have a new custom error for this case, and if `getIdentityType` function throws an instance of that new error, `sortIdentities` function won't throw the error and won't break the app anymore. 

A new IdentityType was also added for Decathlon login. 

## Type of change

## How Has This Been Tested?

A new unit test has been added to check if the fix is effective.

## Checklist:

### DIFF
--- a/tests/unit/sortIdentities.test.ts
+++ b/tests/unit/sortIdentities.test.ts
@@ -462,4 +462,45 @@ describe("sortIdentities", () => {
     const call = sortIdentities(mockedResponse);
     expect(call).toStrictEqual(mockedSortedResponse);
   });
+
+  it("should not throw error if it is an instance of UnhandledIdentityType error when given an unknown identity type", () => {
+    const mockedResponse: { data: { provider: ProviderUser } } = {
+      data: {
+        provider: {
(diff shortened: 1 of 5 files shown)


Fuzznimp (approved):

tests/unit/sortIdentities.test.ts
```diff
@@ -462,4 +462,45 @@ describe("sortIdentities", () => {
...
+                primary: true,
+                status: "validated",
+                type: IdentityTypes.GOOGLE,
+                value: "test@test.test",
+              },
+              {
+                id: "5z8d168a-3f65-4636-9acb-7720a212543r",
+                primary: false,
+                status: "validated",
+                type: "gitlab" as IdentityTypes,
```
Fenntasy, line 493: Just a warning about this kind of things, I think it's better to be explicit with something like `"thisisnotanidentitytype" as IdentityTypes` because we could well implement gitlab in the future 😅

PR #160, opened 2020-12-14
Adding a new identity & marking it as primary directly

Oluane:
## Description

This PR adds the possibility for user to add a new identity and marking it as primary directly. 
To do so, a checkbox input was added beneath identity input in the form. 
A new field `primary: boolean` was added to `InMemoryTemporaryIdentity` & `TemporaryIdentity` which is added in our persistence database, this new field reflects if a new identity needs to be `markIdentityAsPrimary` after its creation on Management side. 

## Type of change

## How Has This Been Tested?

Updated exisiting test suit, and added new assertions to check presence of the new checkbox in the form.

## Checklist:

### DIFF
--- a/src/pages/api/auth-connect/verify-validation-code.ts
+++ b/src/pages/api/auth-connect/verify-validation-code.ts
@@ -5,8 +5,9 @@ import { markIdentityAsPrimary } from "@lib/commands/markIdentityAsPrimary";
 import { checkVerificationCode } from "@lib/queries/checkVerificationCode";
 import { UserCookie } from "@src/@types/UserCookie";
 import type { ExtendedRequest } from "@src/@types/core/ExtendedRequest";
+import { NoIdentityAdded } from "@src/clientErrors";
 import { addIdentityToUser } from "@src/commands/addIdentityToUser";
-import { TemporaryIdentityExpired } from "@src/errors";
+import { GraphqlErrors, TemporaryIdentityExpired } from "@src/errors";
 import { withAuth } from "@src/middlewares/withAuth";
 import { withLogger } from "@src/middlewares/withLogger";
 import { withSentry } from "@src/middlewares/withSentry";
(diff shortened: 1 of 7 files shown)


src/pages/api/auth-connect/verify-validation-code.ts
```diff
@@ -41,7 +42,12 @@ const handler: Handler = async (request: ExtendedRequest, response) => {
             value,
           };
 
-          await addIdentityToUser(body);
+          const newIdentity = await addIdentityToUser(body);
+
+          if (primary) {
+            const { id: identityId } = newIdentity?.data?.addIdentityToUser;
+            await markIdentityAsPrimary(identityId);
+          }
```
Fuzznimp, line 50: What is if `newIdentity` is undefined (i.e. `addIdentityToUser` fails)? Same with `markIdentityAsPrimary`.

Oluane: Since there wasn't any error handling on `addIdentityToUser` before my changes, I assumed that was part of the new errors handling plan you're working on. So I didn't had any checks for that. Was I wrong ?

Fuzznimp: What I'm working on concerns `withAuth` middleware :).
We don't handle errors from `request.session.get` and `getDBUserFromSub` in this file because they are done in `withAuth`, and we haven't found an elegant way to reflect it in the handler yet.

`addIdentityToUser` returns `{data, errors}`. Both can be undefined. You need to react accordingly to the truth of these two objects. Same goes for `markIdentityAsPrimary` :). Let me know if it's unclear.

That's also why we try to avoid using `?` when accessing object's keys, because it basically tells "don't handle the exception if one is thrown".

Oluane: Alright, I think I get it. Working on it ;)

src/pages/api/auth-connect/verify-validation-code.ts
```diff
@@ -41,7 +43,27 @@ const handler: Handler = async (request: ExtendedRequest, response) => {
...
 
-          await addIdentityToUser(body);
+          const identityId = await addIdentityToUser(body).then(
+            ({ errors, data }) => {
+              if (errors) {
+                throw new GraphqlErrors(errors);
+              }
+
+              if (!data) {
+                throw new NoIdentityCreated();
```
Fuzznimp, line 53: What about `NoIdentityAdded`, to stay in line with `addIdentityToUser`?

Fuzznimp (approved):

PR #179, opened 2020-12-23
Deploy to Heroku

Oluane:
## Description

This PR sets up Heroku as the target PaaS for Connect Account. This change was motivated because of the multiple issues we've faced with Vercel (performance) and Render (lack of CI/CD features and performance too even if less worse than Vercel).

Post deploy script: 

To run our script that adds review url to Management on Heroku `postdeploy`, a new `environments` key was added to `app.json`. It precises a different `NODE_ENV` (i.e. `development`) value for `review` environment, since Heroku prunes `devDependencies` in `production` environment, and that these dependencies are needed to run the script (`ts-node`, `typescript` and so on) 

## Type of change

## How Has This Been Tested?

You can checkout the app running [here](https://fwl-connect-deploy-to-h-t4fiya.herokuapp.com)

## Checklist:


app.json
```diff
@@ -0,0 +1,63 @@
...
+            "CONNECT_MANAGEMENT_URL": "https://management.connect.review.fewlines.tech/graphql",
+            "CONNECT_OPEN_ID_CONFIGURATION_URL": "https://fewlines.connect.prod.fewlines.tech/.well-known/openid-configuration",
+            "CONNECT_PROVIDER_URL": "https://fewlines.connect.prod.fewlines.tech",
+            "CONNECT_REDIRECT_URI": "https://account.fewlines.tech/api/oauth/callback",
+            "DYNAMODB_ENDPOINT": "https://dynamodb.eu-west-3.amazonaws.com",
+            "DYNAMODB_REGION": "eu-west-3",
+            "DYNAMODB_TABLE_NAME": "fewlines-connect-account-review",
+            "IS_JWE_SIGNED": "true",
+            "NEXT_PUBLIC_SENTRY_DSN": "https://6baf0213093a428d9e9ec7ae0979868c@o287463.ingest.sentry.io/5387583",
+            "NEXT_TELEMETRY_DISABLED": "true"
```
Fenntasy, line 44: If we don't repeat those, won't it take the global `"env"` ones?

bin/e2e/add-heroku-redirect-uri-to-connect.ts
```diff
@@ -0,0 +1,50 @@
...
+      throw new Error("Connect Application not found");
+    }
+
+    return data.provider.application;
+  });
+
+  const herokuDeploymentUrl =
+    "https://" +
+    process.env.HEROKU_APP_NAME +
+    ".herokuapp.com/api/oauth/callback";
```
Fenntasy, line 32:
```suggestion
    ".herokuapp.com" +
    "/api/oauth/callback";
```

app.json
Oluane: Nope, I thought it would be that, but instead, it totally overrides the "base" ones, and sets just the ones that are beneath the current environment (`environments > review > env`).  cf -> https://devcenter.heroku.com/articles/app-json-schema#environments

Fenntasy: @tgautier do you have any insight to avoid this or is it not avoidable? 😇

tgautier: Even though I totally understand why this change was made, I think that we should keep in mind that the goal of having a CI/CD pipeline and review environments is to test the application behaviour in a production context before it actually reaches the real production.

For me, one of the key tenants of this is to make sure that the reviews, staging and production environments are as similar as possible to each other. I even think we should aim for them being strictly equivalent, except for some values (ENV variables, CPU/RAM resource allocation, etc...).

So, with this in mind, the `NODE_ENV=dev` is actually a big no-no for me, even though I totally understand why this change is suggested.

I would rather have us duplicate the code logic so that the script can be in vanilla JS and the code it was copied from can stay in TS in the actual app.

One of the nice benefits of this would also be that we no longer have to copy all those pesky ENV variables.

tgautier (approved):

PR #187, opened 2021-01-05
React testing library + first test revamp

Oluane:
## Description

The goal of this PR is to add `react-testing-library` package for future replacement of `enzyme` for our components tests. New dependencies has been added (*cf below*), config file was made to customize `render` function with `Next/Router` Context Provider and a first rewrite of an existing test was also made (`HomePage.test.tsx`)

## Type of change

## List of added dependencies

* **@testing-library/react**, v11.2.2
* **@testing-library/jest-dom**, v5.11.8

## Checklist:

### DIFF
--- a/tests/config/testing-library-config.tsx
+++ b/tests/config/testing-library-config.tsx
@@ -1,10 +1,36 @@
 import { render, RenderOptions, RenderResult } from "@testing-library/react";
+import { RouterContext } from "next/dist/next-server/lib/router-context";
+import { NextRouter } from "next/router";
 import React from "react";
 
 import { AccountApp } from "@src/pages/_app.tsx";
 
+const mockedNextRouter: NextRouter = {
+  basePath: "",
+  pathname: "/",
+  route: "/",
+  asPath: "/",
+  query: {},
+  push: jest.fn(),
+  replace: jest.fn(),
+  reload: jest.fn(),
+  back: jest.fn(),
+  prefetch: jest.fn().mockResolvedValue(undefined),
+  beforePopState: jest.fn(),
+  events: {
+    on: jest.fn(),
+    off: jest.fn(),
+    emit: jest.fn(),
+  },
+  isFallback: false,
+};
+
 const AllTheProviders: React.ComponentType = ({ children }) => {
-  return <AccountApp>{children}</AccountApp>;
+  return (
+    <RouterContext.Provider value={{ ...mockedNextRouter }}>
+      <AccountApp>{children}</AccountApp>
+    </RouterContext.Provider>
+  );
 };
 
 const customRender = (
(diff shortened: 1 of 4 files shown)


tests/config/testing-library-config.tsx
```diff
@@ -0,0 +1,45 @@
...
+const customRender = (
+  ui: React.ReactElement,
+  options?: Omit<RenderOptions, "queries">,
+): RenderResult => render(ui, { wrapper: AllTheProviders, ...options });
+
+// re-export everything
+export * from "@testing-library/react";
+
+// override render method
+export { customRender as render };
```
Fuzznimp, line 45:
```suggestion
export * from "@testing-library/react";
export { customRender as render };
```

What do you think? De we really need the comments here?

tests/pages/HomePage.test.tsx
```diff
@@ -1,24 +1,14 @@
-import { mount } from "enzyme";
```
Fuzznimp, line 1: Maybe add a check for the background?

Fuzznimp (commented):
It's probably too soon to judge, but any ouputs so far on `testing-libraries` vs `enzyme`?

tests/pages/HomePage.test.tsx
Oluane: I spent a lot of time this morning trying to figuring out how we could test that, because it's not really a background (in term of css) but a svg that takes full width.
Then, I read some interesting things about components testing and the philosophy behind `testing-library` (cf [here](https://kentcdodds.com/blog/testing-implementation-details) ) and I ended up thinking that testing this bg presence was more about implementation details. From that, I decided to test if there was the provider name & the link to connect displayed, since this is the kind of thing the user will see(k). This reasoning is also motivated by the fact that these two points were less change-sensitive and would cause less test refacto, instead of the BG implementation (some client could decide to put this bg using css and not a svg). I hope my thought seems clear to you 😅

tests/config/testing-library-config.tsx
Oluane: I think we don't 😳

tests/pages/HomePage.test.tsx
Fuzznimp: That's a very good point :). Next question then:
Do we need this test then? We are already testing the link in the e2e tests, so all it really does it to check if we have `Fewlines` written.

Oluane: I don't have answer for this question right now. The limit is hard to find. In this case, almost every component tests is useless, so I don't know what to think.

Fuzznimp: We probably should have a discussion about this then. There is no real answer, we just need to decide if we heavily rely on e2e tests, on split them with unit/integration tests.

Fuzznimp (approved):

PR #189, opened 2021-01-06
Revamp ValidateIdentityPage test with RTL

Oluane:
## Description

This PR's goal was to rewrite `ValidateIdentityPage.test.tsx` to use `react-testing-library` instead of `enzyme`. Most of previous test cases have been kept, and some new added.

## Type of change

## Issues

error screenshot

During the rewriting of the test suit (ie `ValidateIdentityPage.test.tsx`), this error rose up. After some tests, it seems to only happen when a minimum of two `describe()` blocks are present in the test file. When any of the two `describe()`blocks are runned alone, this error doesn't occur. 

As the error suggested, this is coming from `next/link` component, that I didn't mock in the first place. I tried different things inside the test suit itself (like running `cleanup()` with `beforeEach()` or `afterEach()` in and outside the `describe()` blocks, calling `render()` inside `act()` ...etc), unfortunately, none solved the issue.

The only workaround that solved this issue was to add a mock of `next/link` inside `testing-library-config.tsx`, where I had already added a mock for Next's Router before this issue. Here is the code :

```typescript
const mockLink = ({
  children,
  ...props
}: {
  children: JSX.Element;
}): JSX.Element => {
  return React.cloneElement(children, props);
};

jest.mock("next/link", () => {
  return mockLink;
});
```
In the first place, I only returned children inside the mock. But since Next's Link is a wrapper for anchor tag (`<Link><a>....</a></Link>`), the children anchor tag wasn't receiving anymore the `href` attribute, which might be an important attribute to test. So I ended up cloning the `children` prop (`<a>`) to be able to pass it the other props of `<Link>` component and to keep its original behaviour. 

## How Has This Been Tested?

Test suit.

## List of added dependencies

* **@testing-library/user-event**, v12.6.0.

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "@storybook/react": "6.1.7",
     "@testing-library/jest-dom": "5.11.8",
     "@testing-library/react": "11.2.2",
+    "@testing-library/user-event": "12.6.0",
     "@types/enzyme": "3.10.8",
     "@types/enzyme-adapter-react-16": "1.0.6",
     "@types/jest": "26.0.15",
--- a/src/components/display/fewlines/Icons/CrossIcon/CrossIcon.tsx
+++ b/src/components/display/fewlines/Icons/CrossIcon/CrossIcon.tsx
@@ -3,6 +3,7 @@ import React from "react";
 export const CrossIcon: React.FC = () => {
   return (
     <svg width="14" height="14" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <title>Closing cross</title>
       <path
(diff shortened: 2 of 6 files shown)


Fuzznimp:
@Fenntasy Pinging you just so you can be aware of the `Issues` section's content.

Oluane:
@Fuzznimp I already did it 😄

Fenntasy:
Yep, aware! I'm curious but I don't see any problem with the current fix. As I said, it would be good to monitor it from time to time.

Fuzznimp (approved):

PR #207, opened 2021-01-15
Creating custom check for e2e tests & updating it according test results

Oluane:
## Description

This PR's goal was to add the creation of a custom `check-run` on each new `push` on a PR, which is related to e2e tests success completion and will serve to determine if the PR can be merged. 

On a first GH Actions workflow file, which is triggered by `pull_request` events, we fetch GH API to create the `check-run` associated with the commit that started the workflow. 

On a second workflow file, triggered by `deployment_status` events, and where the e2e tests are actually runned, we use their results or the `deployment` state (success/failure) to determine the final status of the custom `check-run`.

## Type of change

## How Has This Been Tested?

Mannually & test suit.

## Checklist:


.github/workflows/e2e-test-heroku.yml
```diff
@@ -57,3 +57,42 @@ jobs:
...
+          path: |
+            ${{ steps.yarn-cache-dir-path.outputs.dir }}
+            **/node_modules
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: dependencies
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn install
```
Oluane, line 91: @Fenntasy I was wondering myself if we really need to checkout the project and install every deps, since in this script (`update-e2e-run-check-status.ts`), we don't use many of the project deps. Couldn't we just `yarn install ...` the deps we need for this script execution and save some time on this job ?

Fenntasy: If `yarn` (or `npm`) worked in the way you think, yes we could absolutely do that.
Unfortunately, `yarn install package` will install everything:

```sh
~/W/f/connect-account ❯ rm -rf node_modules
~/W/f/connect-account ❯ yarn add jest
yarn add v1.22.5
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
[4/5] 🔗  Linking dependencies...
[5/5] 🔨  Building fresh packages...
success Saved lockfile.
warning "jest" is already in "devDependencies". Please remove existing entry first before adding it to "dependencies".
success Saved 1 new dependency.
info Direct dependencies
└─ jest@26.6.3
info All dependencies
└─ jest@26.6.3
✨  Done in 35.94s.
~/W/f/connect-account ❯ yarn
yarn install v1.22.5
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.57s.
```

Oluane: I see, but what if if we don't execute this step inside the project ? (I mean, without that `- uses: actions/checkout@v2`). There won't be any `package.json` so we possibly could install just what we need, no ? 🤔  
(I'm not saying that we absolutely need to do that, but it's a question that I had when working on this task and I wanted your thought about it)

Fenntasy: How would you get the script then?

Oluane: Yep, just after writing the last comment, I realised that. Forget about what I said 😅

Fenntasy: I didn't mean it like that, it would be totally possible to fetch only the script we need (via github blog urls) for instance.
But that would be harder to deal with when we're updating the script.

.github/workflows/lint-and-test.yml
```diff
@@ -57,4 +57,33 @@ jobs:
...
+    
+      - name: dependencies
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn install
+
+      - name: creating e2e tests check-run
+        env:
+          GITHUB_CONTEXT_EVENT: ${{ toJson(github.event) }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: yarn e2e:create-check-run
```
Fenntasy, line 89:
```suggestion
        run: yarn e2e:create-check-run

```
😁

.github/workflows/e2e-test-heroku.yml
Fenntasy: Thinking out loud, one way of doing that would be to have another `package.json` in the `bin` folder and consider it totally separate from the rest of the application (that would force us to not use the rest of the code though). But it would be a considerably smaller installation

bin/e2e/create-e2e-check-run.ts
```diff
@@ -0,0 +1,33 @@
...
+    {
+      method: "POST",
+      headers: {
+        Authorization: `Bearer ${process.env.GITHUB_TOKEN}`,
+      },
+      body: JSON.stringify(checkRunBody),
+    },
+  ).catch((error) => {
+    throw error;
+  });
```
Fenntasy, line 30: I think this is the same as not catching it in the end, no?

bin/e2e/update-e2e-check-run-status.ts
```diff
@@ -0,0 +1,57 @@
...
+  const githubActionsContext = JSON.parse(process.env.GITHUB_CONTEXT_EVENT);
+
+  if (process.env.GITHUB_TOKEN === undefined) {
+    throw new Error("GITHUB_TOKEN environment variable is undefined");
+  }
+
+  const hasDeploymentFailed =
+    githubActionsContext.deployment_status.state === "failure";
+
+  const checkSuiteList = await fetch(
```
Fenntasy, line 17: If the deployment has failed, do we need to get the checks or can we immediately update it to failure?

Oluane: I think we need to get the checks anyway, because we need to retrieve the custom check id to update its status (to construct the URL on which we're going to do the PATCH thing). 

In fact, I spend some time on thinking how to "pass" this id between workflows (`lint-and-test.yml` is where we have it in the create check run script) but I concluded it will be more simple to fetch all checks (since we need to do it to retrieve the one concerning real e2e tests if deployment is ok anyway)

Fenntasy: Forgot about that, you're right 👍

.github/workflows/lint-and-test.yml
Oluane: I always get stuck with EOF and YML files (VS Code doesn't do it for me) 😅

bin/e2e/create-e2e-check-run.ts
Oluane: I guess ? In fact, I'm not sure of anything on this point. I will try without the catch 🙂

.github/workflows/e2e-test-heroku.yml
Oluane: Hmm I don't really see how to do that. Does this action `actions/checkout@v2` have some options to checkout more "deeper" in the project ? Or would it be more like giving volumes to a docker image ?

Fenntasy: Well, first, it's not something I would suggest to do in this PR but in the future.

But no, I don't know if we can make a deep checkout but that action is around 2 seconds I think so that's really not the problem. The long part is the get-cache/yarn-install/set-cache and that could be done faster

Oluane: Yep, I totally agree, it's not something worth doing now. But I would like to discuss it one day, if we decide to go that way 🙂

Fenntasy (approved):

PR #215, opened 2021-01-25
First implementation `@fwl/web`

Fuzznimp:
## Description

This PR aims at implementing `@fwl/web` for 2 paths:
- (API) `/api/delete-identity`
- (SSR) `/`

## Type of change

## How Has This Been Tested?

Manually and with the tests suit.

## Screenshot

image

## List of added dependencies

- `@fwl/tracing`: "0.7.0",
- `@fwl/web`: "0.6.2",

## Checklist:


Fuzznimp:
@Fenntasy @Oluane Couples things:
- I couldn't find any relevant attributes to add to the API handler's span.
- Am I missing something, or is it not possible to set span attributes inside the `getServerSidePropsWithMiddlewares`?

src/tracer.ts
```diff
@@ -0,0 +1,12 @@
+import { getTracer, startTracer, Tracer } from "@fwl/tracing";
+
+startTracer({
+  simpleCollector: {
+    serviceName: "connect-account",
+    url: "http://localhost:9411/api/v2/spans",
```
Fenntasy, line 6: both of these should come from the config

Fenntasy: sorry, the whole `simpleCollector` should come from the config :/

Fuzznimp: b74e847

Looks like we can only use public Satellites with Lightstep, so here is my take. I will document it if this is good enough.

src/tracer.ts
```diff
@@ -1,11 +1,23 @@
...
-  simpleCollector: {
-    serviceName: "connect-account",
-    url: "http://localhost:9411/api/v2/spans",
-  },
-});
+import { config } from "@src/config";
+
+const options = config.lightstepAccessToken
+  ? {
+      serviceName: config.tracerServiceName,
```
Fenntasy, line 7: you can get rid of this line but otherwise, 👍

Fuzznimp: 1a34baf
Ah you are right :).

.env_vars.sample
```diff
@@ -26,6 +26,10 @@ export CONNECT_JWT_ALGORITHM=***
 export ACCOUNT_JWE_PRIVATE_KEY=***
 export IS_JWE_SIGNED=false
 
+## Tracer
+export TRACER_SERVICE_NAME=***
+export LIGHTSTEP_ACCESS_TOKEN=***
```
Fenntasy, line 31:
```suggestion
export TRACER_SERVICE_NAME=
export LIGHTSTEP_ACCESS_TOKEN=
```

especially the second one since, if we copy the sample, it will ends up using the token as `***` is not empty 😅

src/logger.ts
```diff
@@ -0,0 +1,6 @@
+import { createLogger, EncoderTypeEnum } from "@fwl/logging";
+
+export const logger = createLogger({
+  service: "connect-account",
```
Fenntasy, line 4: Should we name `TRACER_SERVICE_NAME` `SERVICE_NAME` and use it both for the tracer and the logs? 😇

Fuzznimp: Smart move.

Fuzznimp: Do you agree with
```ts
config.serviceName = process.env.SERVICE_NAME || "Connect.Account";
```

Oluane (dismissed):

src/logger.ts
Fenntasy: yes

Oluane (dismissed):

Fenntasy (approved):

Oluane (approved):

PR #233, opened 2021-02-10
First implementation of update identity flow

Oluane:
## Description

This PR is a first draft of the update identity flow. The goal was to use a maximum of existing workflows to assemble an update one. 

So the flow starts like a basic add identity one, where we send a validation code, we then redirect to the validation code form and check the validity of the new identity before sending it to Management. 

To compele this flow, just after sending the new identity to Management, we delete the ancient one. To do so, I added an optional key (`identityToUpdateId`) to the `TemporaryIdentity` object that we store in our user table in DynamoDB. This key value is the id of the identity the user wants to update, and we use it to retrieve the old identity value and delete it. 

## Type of change

## Checklist:


.env_vars.sample
```diff
@@ -32,6 +32,7 @@ export LIGHTSTEP_ACCESS_TOKEN=
 
 ## Sentry
 export NEXT_PUBLIC_SENTRY_DSN=https://6baf0213093a428d9e9ec7ae0979868c@o287463.ingest.sentry.io/5387583
+export SENTRY_ENVIRONMENT=
```
Fenntasy, line 35:
```suggestion
export SENTRY_ENVIRONMENT= development
```

no?

src/pages/api/auth-connect/send-identity-validation-code.ts
```diff
@@ -55,16 +56,24 @@ const handler: Handler = (request, response): Promise<void> => {
       })
         .then(async ({ eventId }) => {
           span.setDisclosedAttribute("is validation code sent", true);
-
-          const temporaryIdentity = {
+          console.log("bqh lol");
```
Fenntasy, line 59: 🚒

src/pages/api/auth-connect/send-identity-validation-code.ts
```diff
@@ -55,16 +56,24 @@ const handler: Handler = (request, response): Promise<void> => {
...
 
+          if (identityToUpdateId) {
+            temporaryIdentity = {
+              ...temporaryIdentity,
+              identityToUpdateId,
+            };
+          }
+
           await insertTemporaryIdentity(userCookie.sub, temporaryIdentity);
+          console.log("rip");
```
Fenntasy, line 76: 🧯

Fenntasy (commented):
Am I correct when thinking this could move the primary identity if we have 2 identies when updating the primary?

src/pages/api/auth-connect/send-identity-validation-code.ts
Fuzznimp: As you can tell by the "q", these were mine..

.env_vars.sample
Fuzznimp: Now that I think about it, the line
```ts
enabled: process.env.NODE_ENV === "production",
```
should prevent any report in dev mode. That's why I did not add the variable in our local `env_vars`.

Oluane:
> Am I correct when thinking this could move the primary identity if we have 2 identies when updating the primary?

I'm not sure to clearly understand what you mean, but I can tell you that if you try to update the primary, the new one will be the new primary in the end.

src/pages/api/auth-connect/send-identity-validation-code.ts
Oluane: Oops 😳

src/pages/api/auth-connect/send-identity-validation-code.ts
Oluane: 😱 🙇

Fenntasy:
> I'm not sure to clearly understand what you mean, but I can tell you that if you try to update the primary, the new one will be the new primary in the end.

Even if you have:
- identity A, primary
- identity B, created in second
- update A to C 

?

Oluane:
> Even if you have:
> 
>     * identity A, primary
> 
>     * identity B, created in second
> 
>     * update A to C
> 
> 
> ?

Yep, when setting the `TemporaryIdentity` object for the new Identity, I use the `primary` value from the one we want to update. So, in `verify-validation-code.ts`, after `addIdentityToUser` command, it triggers `markIdentityAsPrimary` command and sets the new as primary. (Management handles on its own the switch between primary).

Fuzznimp (approved):

PR #238, opened 2021-02-16
Implement webErrors on API pages

Oluane:
## Description

This PR implements the new webErrors on all API pages, with a lot of refacto concerning errors handling. I may have missed some things, feel free to point it up. 

I voluntarily did not implement them on all commands concerning DynamoDB. I wasn't sure what the move was, so I'm waiting your opinion/thoughts about it. 

## Type of change

- [x] **Breaking change** (fix or feature that would cause existing functionality to not work as expected).

## Checklist:


src/pages/api/auth-connect/mark-identity-as-primary.ts
```diff
@@ -42,27 +56,37 @@ const handler: Handler = async (request, response) => {
...
+              throw webErrorFactory(webErrors.identityNotFound);
+            }
+
+            if (error instanceof ConnectUnreachableError) {
+              span.setDisclosedAttribute("Connect unreachable", error.message);
+              throw webErrorFactory(webErrors.connectUnreachable);
+            }
+
+            span.setDisclosedAttribute("is Identity marked as primary", false);
+            throw error;
```
Fenntasy, line 80: I have questions:
- should the `span.setDisclosedAttribute("is Identity marked as primary", false);` be at the top if we want it to be on traces even if we throw an error?
- shouldn't we have the same naming for an attribute that is an error message? to facilitate looking for it? 

The recovery middleware uses:
```

          span.setDisclosedAttribute("exception.class", error.toString());
          span.setDisclosedAttribute("exception.message", error.message);
          span.setDisclosedAttribute("stack_trace_hash", error.stack);
          ```

File: tests/unit/validate-identity.test.ts [hunk 2]
```diff
@@ -18,7 +18,7 @@ describe("validateIdentity unit test", () => {
     expect.assertions(2);
 
     const mockedFetchResponse = new Response(
-      JSON.stringify({ error: "INVALID" }),
+      JSON.stringify({ code: "invalid_validation_code" }),
```
Fenntasy, line 21: what about importing them from the errors module?

Oluane: Good point

src/pages/api/auth-connect/mark-identity-as-primary.ts
Oluane: For the 1st point, I totally agree, these kind of spans should be on top.

And for the second one, it might be useful indeed if we can filter by span names. So we go for `exception.message` when we send the error message ?

src/pages/api/auth-connect/send-identity-validation-code.ts
```diff
@@ -29,100 +30,120 @@ import { logger } from "@src/logger";
...
-            temporaryIdentity = {
-              ...temporaryIdentity,
-              identityToUpdateId,
+  const webErrors = {
+    badRequest: ERRORS_DATA.BAD_REQUEST,
+    identityInputCantBeBLank: ERRORS_DATA.IDENTITY_INPUT_CANT_BE_BLANK,
+    temporaryIdentityNotFound: ERRORS_DATA.TEMPORARY_IDENTITY_NOT_FOUND,
+    temporaryIdentitiesNotFound: ERRORS_DATA.TEMPORARIES_IDENTITY_NOT_FOUND,
+    connectUnreachable: ERRORS_DATA.CONNECT_UNREACHABLE,
+  };
```
Fuzznimp:
```suggestion
  const webErrors = {
    badRequest: ERRORS_DATA.BAD_REQUEST,
    identityInputCantBeBlank: ERRORS_DATA.IDENTITY_INPUT_CANT_BE_BLANK,
    connectUnreachable: ERRORS_DATA.CONNECT_UNREACHABLE,
  };
```

Fuzznimp: In case you miss it, there is a type on `Blank` (`BLank` in your code).

src/pages/api/auth-connect/mark-identity-as-primary.ts
Fuzznimp: We could create a function that takes the span, the error and add the attributes to the span.

Fenntasy: > We could create a function that takes the span, the error and add the attributes to the span.

Well the error middleware does that, but it doesn't unwrap the error that made us use the weberror (maybe something to deal with in the future)

Fuzznimp: What do you mean by `unwrap`?

Fenntasy:
```
if (error instanceof ConnectUnreachableError) {
  span.setDisclosedAttribute("Connect unreachable", error.message);
  throw webErrorFactory(webErrors.connectUnreachable);
}
```
in this example, the error middleware will only see the webError, it won't have access to the `ConnectUnreachableError` . You could do something like `throw webErrorFactory(webErrors.connectUnreachable, error)` and create a webError that uses the `parentError` option (I think we have one). But I don't think we use this in the error middleware so we display the WebError but not what has caused the WebError. So we would need to 'unwrap' it from the WebError.

Dunno if I'm clearer though

src/pages/api/auth-connect/verify-validation-code.ts
```diff
@@ -2,9 +2,11 @@ import {
   addIdentityToUser,
```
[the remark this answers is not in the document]
Fenntasy: uuuh I don't get your point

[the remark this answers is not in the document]
Fuzznimp: Well, instead of having 4 `.catch()` that do the same thing, i.e.
```ts
foo().catch(
  if (error instanceof GraphqlErrors) {
    throw webErrorFactory(webErrors.identityNotFound);
  }

  if (error instanceof ConnectUnreachableError) {
    throw webErrorFactory(webErrors.connectUnreachable);
  }

  throw error;
);

bar().catch(
  if (error instanceof GraphqlErrors) {
    throw webErrorFactory(webErrors.identityNotFound);
  }

  if (error instanceof ConnectUnreachableError) {
    throw webErrorFactory(webErrors.connectUnreachable);
  }

  throw error;
);

jean().catch(
  if (error instanceof GraphqlErrors) {
    throw webErrorFactory(webErrors.identityNotFound);
  }

  if (error instanceof ConnectUnreachableError) {
    throw webErrorFactory(webErrors.connectUnreachable);
  }

  throw error;
);

michel().catch(
  if (error instanceof GraphqlErrors) {
    throw webErrorFactory(webErrors.identityNotFound);
  }

  if (error instanceof ConnectUnreachableError) {
    throw webErrorFactory(webErrors.connectUnreachable);
  }

  throw error;
);
```

we would have

```ts
try {
  foo();
  bar();
  jean();
  michel();
} catch (error) {
 if (error instanceof GraphqlErrors) {
    throw webErrorFactory(webErrors.identityNotFound);
  }

  if (error instanceof ConnectUnreachableError) {
    throw webErrorFactory(webErrors.connectUnreachable);
  }

  throw error;
}
```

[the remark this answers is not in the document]
Fenntasy: ah, yes, didn't see it in my review 😅 
I agree

src/pages/api/auth-connect/mark-identity-as-primary.ts
```diff
@@ -42,27 +56,38 @@ const handler: Handler = async (request, response) => {
...
 
       span.setDisclosedAttribute(
         "is identity marked as primary authorized",
         false,
       );
 
-      response.statusCode = HttpStatus.BAD_REQUEST;
-      response.end();
-      return;
+      throw webErrorFactory(webErrors.badRequest);
```
Fenntasy, line 90: What would you think about reversing this `if`?
Meaning having the bad path happening before the happy path:

```
if (!isAuthorized) {
  span.setDisclosedAttribute(
    "is identity marked as primary authorized",
    false,
  );
  throw webErrorFactory(webErrors.badRequest);
}

// rest of the code
```

I feel this would be more readable, what do you think? (ping @Fuzznimp )

Fuzznimp: I'm all for it. I've started to do this, and in most cases it's way more clean.

Oluane: Good for me too 👍

src/pages/api/auth-connect/mark-identity-as-primary.ts
Fuzznimp: It Is indeed way clearer. Thanks.

We would need to add a check in the error middleware thought. Something that check that, if the error catches by the middleware has a `parentError`, it uses it to compose the span instead of the error catch.

src/pages/api/auth-connect/verify-validation-code.ts
```diff
@@ -42,6 +57,8 @@ const handler: Handler = async (request, response) => {
     const user = await getDBUserFromSub(userCookie.sub);
```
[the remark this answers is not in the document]
Oluane: Do you have any idea in mind concerning the kind of errors that should throw the DynamoDB commands ?

src/pages/api/auth-connect/verify-validation-code.ts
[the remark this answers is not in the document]
Oluane: This one will be done on another PR, we need to improve our DB command errors before changing to try/catch.

Fuzznimp (dismissed):

src/pages/api/auth-connect/send-identity-validation-code.ts
```diff
@@ -87,38 +109,32 @@ const handler: Handler = (request, response): Promise<void> => {
...
-            );
-
-            response.statusCode = HttpStatus.BAD_REQUEST;
-            response.json({ errorMessage: error.message });
-            return;
+          span.setDisclosedAttribute("is validation code sent", false);
+
+          if (error instanceof IdentityAlreadyUsedError) {
+            throw webErrorFactory(webErrors.badRequest);
           }
```
Fuzznimp, line 116: I was thinking about task `CU-egxugf`. At the moment, Sentry reports exception raised in case of a wrong input, which is an issue. We need to have a way to recognise this error inside the `withAuth` middleware, without telling explicitly that the Identity inputed is already taken.

I can think of two solutions, one expensive, one sub-optimal.

The first one is to add more abstraction to our error's codes, something like what we have in `sparta-api`. This is expensive because we'll need to change all our current web errors, and we all agree that the error's codes in `sparta-api` are probably not the best solution.

The second one would be to add an error content value in the web error, with something like a code. I don't like this one because it would make the algorithm part in `with-auth` obnoxious.

What do you guys think?

Fuzznimp: I guess we could do this in an other PR.

Fuzznimp (approved):

PR #247, opened 2021-02-25
Handling empty password input 

Oluane:
## Description

This PR adds a handling of trying to update a password with empty value. Now, if user tries to do it, we display an error message, just as we already did for the add identity flow. The way it is implemented, no query is made to Management, we handle this case inside Account. 

This PR is a fix for this Sentry [error](https://sentry.io/organizations/fewlines/issues/2235288539/?project=5387583&referrer=slack)

## Type of change

## Screenshots

image

## Checklist:


src/workflows/set-password.ts
```diff
@@ -9,11 +9,16 @@ type SetPasswordOutput = {
...
+  passwordInput: string,
+): Promise<
+  | { isUpdated: boolean }
+  | { code: string; message: string; details?: PasswordRules }
+> {
   return fetchJson("/api/auth-connect/set-password", HttpVerbs.POST, {
     passwordInput,
   }).then((response) => {
-    if (response.status >= 400 && response.status !== 422) {
+    if (response.status > 400 && response.status !== 422) {
```
Fuzznimp, line 21: Why not include `400`s?

Oluane: I excluded 400s because if I don't, the invalidBody webError get catch there and triggers an unhandled error (the one we had Sentry report yesterday). I followed what we did in `addIdentity` workflow : 

```typescript
if (response.status > 400) {
      throw new ErrorSendingValidationCode();
}
```

But I really wonder if we should keep these code blocks in the workflows. If the `ConnectUnreachable` webError is raised, it will get catched inside them too.

Fuzznimp: I agree we should remove this workflow 👍

Fuzznimp (approved):

PR #340, opened 2021-03-25
Verify two FA validation code

Oluane:
## Description

This PR adds the two fa validation code handler, and improve the input style which now looks like a 6 digit inputs. 

Regarding handler side: 

* The validation code verification process handles multiple `sudo_event_ids` attempts.
* For now, in the handler call on verification code form submit, we handle both empty input value and wrong code input, which triggers the display of an error message in the UI. If no valid `sudo_event_ids` is found, we send the user back to step one (aka contact identity choice to send validation code).
* Added a new Dynamo command `removeExpiredSudoEventIds` which takes as argument an array of valid `sudo_event_ids` and replaces the previous one in the dynamo db user.

Regarding input style: 

* To make the text input for validation code looks like a 6-digits input, and for accessibility purpose, we still use the accessible component `InputText`. It just don't have any style apart from its displayed value. The 6-digits input is just a set of `spans` elements superposed on original input and is only made with CSS. To do so, I had to modify a bit `InputText` for it to accept children, it nows dynamically render the input with its children if `children` props is provided or the casual input otherwise. Note : the original input masking and the `children` style must be defined where it will be used for now. 
* I had also to deactivate the default browser focus style on the original input but reproduce it on the `spans`. (no support on IE).

I added a first draft of integration test for the whole page (`SudoPage`), which will maybe need to be enhanced at some point later. 

## Type of change

## Screenshots

image

## Checklist:


Fenntasy:
Task linked: [CU-g57826 Rework of code verification form + verification code handler](https://app.clickup.com/t/g57826)

src/components/forms/verify-two-fa-code-form.tsx
```diff
@@ -1,31 +1,80 @@
...
+                parsedResponse.message ===
+                ERRORS_DATA.SUDO_EVENT_IDS_NOT_FOUND.message
+              ) {
+                setFormID(uuidv4());
+                setIsCodeSent(false);
+              }
+            }
+
+            // router && router.push("/account/security/update");
+          });
```
Fuzznimp, line 57: What about using this here https://github.com/fewlinesco/connect-account/blob/d79e4d33d1eabd576ac6a239c25212d0521a7cdd/src/components/forms/validate-identity-form.tsx
To uniform the way we do it. What do you think?

src/pages/account/security/sudo.tsx
```diff
@@ -47,63 +56,63 @@ const getServerSideProps: GetServerSideProps = async (context) => {
...
-      //   if (primaryIdentities.length > 1) {
-      //     return {
-      //       props: {
-      //         primaryIdentities,
-      //       },
-      //     };
-      //   }
-      // }
+        if (primaryIdentities.length > 1) {
+          return {
```
Fuzznimp, line 107: Shouldn't this be `> 0`? We can have a case where the user only have 1 primary identity.

src/pages/api/auth-connect/verify-two-fa-validation-code.ts
```diff
@@ -0,0 +1,190 @@
...
+
+      if (!verificationCode) {
+        throw webErrorFactory(webErrors.invalidBody);
+      }
+
+      const userCookie = (await getServerSideCookies<UserCookie>(request, {
+        cookieName: "user-cookie",
+        isCookieSealed: true,
+        cookieSalt: config.cookieSalt,
+      })) as UserCookie;
```
Fuzznimp, line 61: Why using a type assertion instead of
```
      if (!userCookie) {
        response.statusCode = HttpStatus.TEMPORARY_REDIRECT;
        response.setHeader("location", "/");
        response.end();
        return;
      }
  ```
  ?

src/components/forms/verify-two-fa-code-form.tsx
Oluane: Absolutely.

src/pages/api/auth-connect/verify-two-fa-validation-code.ts
Oluane: Well, I should change it too in `verify-validation-code` then.

Fuzznimp: Hm, I still dunno which one to choose tbh.

Oluane: To uniform all handlers, I think we should put this code snippet in the two verification handlers that had not it yet and then we'll see what we do, what do you think ?

Fuzznimp (approved):

Fuzznimp (approved):

PR #434, opened 2021-04-09
Fix 500 on wrong identity ID slug

Fuzznimp:
## Description

This PR aims at fixing an issue when changing the identity ID from the URL slug linking to a resource not found.

I had to update `fewlines/connect-management` to raise an `IdentityNotFoundError` exception.

I had to fix the 404 fallback component spacing in through the layout. To help with it, I finalised the trasition of the `breadcrumbs` props to only be `string`.

image

I have tested my change for both `[id]` and `[id]/update`, which are the two routes who calls `/api/identities/get-identity`.

## Type of change

## Checklist


Fenntasy:
Task linked: [CU-gt7ym8 [investigate] - Errors/Web-errors throwns in API handlers are not catched by `useSWR`](https://app.clickup.com/t/gt7ym8)

src/pages/api/identity/get-identity.ts
```diff
@@ -56,13 +58,20 @@ const handler: Handler = async (request, response) => {
     }).catch((error) => {
       span.setDisclosedAttribute("is Identity fetched", false);
 
-      if (error instanceof GraphqlErrors) {
+      if (error instanceof IdentityNotFoundError) {
         throw webErrorFactory({
           ...webErrors.identityNotFound,
           parentError: error,
```
Oluane, line 64: I think we don't need to pass the `parentError` in this case, since it's a waited and known exception, the concatened messages of both errors will be duplicates in the end. What do you think ?

Fuzznimp: That's actually a good point. It will make the error message way cleaner 👍 .

Oluane (approved):

PR #471, opened 2021-04-22
Fix CSS heritage and BoxedLink extension

Fuzznimp:
## Description

This PR aims at cleaning up some CSS heritage, component extension and forced dimensions.

In the last file, the code's indentation from GitHub is messed up:
image
But if you check the file, you'll see it's a preview issue:
image

Not sure what is causing that. 

## Type of change

## Checklist

- [ ] My changes generate no new warnings.


Fenntasy:
Task linked: [CU-hk55j7 [fix] - CSS heritage to prevent conflicts inside `profile-overview`](https://app.clickup.com/t/hk55j7)

Oluane (commented):
👍

src/components/skeletons/skeletons.tsx
```diff
@@ -1,74 +1,5 @@
...
-  border-radius: ${({ theme }) => theme.radii[1]};
-
-  @keyframes loading {
-    100% {
-      transform: translateX(100%);
-    }
-  }
-`;
-
 const SkeletonAnimation = styled.div`
```
Fenntasy, line 3: Nothing there is used anymore? :o

src/components/badges/badges.tsx
```diff
@@ -48,12 +43,11 @@ const Wrapper = styled.div`
...
-  margin-bottom: 1.5rem;
+`;
 
-  p {
-    font-size: ${({ theme }) => theme.fontSizes.xs};
-    margin-right: 0.3rem;
-  }
+const BadgeLabel = styled.p`
+  font-size: ${({ theme }) => theme.fontSizes.xs};
+  margin-right: 0.3rem;
```
Fenntasy, line 50: Let's put a little warning there, I don't think there should be random `margin-right` ^^

Fenntasy (approved):

src/components/skeletons/skeletons.tsx
Fuzznimp: Nop, as we only load text now, or in special cases, we re-create a disabled `<BoxedLink />`.

src/components/badges/badges.tsx
Fuzznimp: I didn't want to make too many changes there (as I thought it was already the case), but after checking a bit of the code base, we have a lot of those. We'll keep improving these. :).

Fenntasy: Yes, it was really just to put a pin in there for the future

Fuzznimp: Also, would you consider useful for us to switch to a grid system? As much I love it, I'm not sure whether it's worth the time to make the migration (because I see it as expensive)

Fenntasy: Maybe a bit too soon for that, let's start by rationalizing before that ^^

PR #491, opened 2021-04-23
Implementation of date picker for updating user-profile

Oluane:
## Description

This PR adds a datepicker input. Since `react-aria` doesn't provide anything related to datepicker, I had to check external lib to avoid reinventing the wheel. So I tested two external lib: 
* `react-datepicker`
* `react-day-picker`

I chose the first one in the end, `react-datepicker`, because it has accessilibity features out-of-the-box. It is heavier than the second one (_189.3kB -> 40.9kB_), but it's a valuable trade-off imho.

Due to the issue we had with mock server on review app, I made some changes on this PR that aren't related to the inital task. I updated `app.json` to add Profile mock server URL, added `profile address` to the existing scopes. And, for our e2e tests suites to work, I had to add a step inside Taiko instructions in `authenticate-to-connect.ts` to accept the new scopes to be able to access the app. 
EDIT : Since we need to do these instructions just once when the full test suit is ran, I added it inside a conditionnal statement on the existance or not of a piece of text present in the scopes acceptance page. I had to reduce the default `retryInterval` of `.exists()` from 100 000ms to 10 000ms, because it was causing time out issues with `delete-identity.test.ts` test, in which we already had to slow Taiko because of SWR. 10 000ms seems to be good for now, but it will maybe cause "random" fails in the future if something in the process takes more time. 

## Type of change

## Issues

The poppin calendar is quite small imo, I tried to manipulate that, but there's no easy way of doing it. Thus it is possible, I would mean a lot of changes and would take a some time to adapt it without losing out-of-the-box responsiveness. 

## List of added dependencies

* react-datepicker: v3.8.0,
* @types/react-datepicker: v3.1.8,

## Screenshots

image
image

## Checklist


Fenntasy:
Task linked: [CU-hf3uvx Implement front form for update profile birthdate](https://app.clickup.com/t/hf3uvx)

src/components/input/input-date-picker.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <DatePicker
+            selected={new Date(selected)}
+            onChange={onChange}
+            dateFormat={"yyyy-MM-dd"}
+          />
+        ) : (
+          <SkeletonTextLine fontSize={1.4} />
+        )}
+      </Label>
+    </>
```
Fuzznimp, line 33: Pretty sure you don't need the `React.Fragment` here :).

src/pages/account/profile/user-profile/update.tsx
```diff
@@ -23,13 +23,11 @@ import { sentryMiddleware } from "@src/middlewares/sentry-middleware";
 const UpdateUserProfilePage: React.FC = () => {
```
Fuzznimp, line 23: Following the recent discussion, the route should be `src/pages/account/profile/user-profile/edit.tsx`, only the form should be named `Update`.

app.json
```diff
@@ -22,7 +22,8 @@
     "NEXT_PUBLIC_SENTRY_DSN": "https://6baf0213093a428d9e9ec7ae0979868c@o287463.ingest.sentry.io/5387583",
     "NEXT_TELEMETRY_DISABLED": "true",
     "SENTRY_ENVIRONMENT": "preview",
-    "NEXT_PUBLIC_FEATURE_FLAG": "false"
+    "NEXT_PUBLIC_FEATURE_FLAG": "false",
+    "PROFILE_MOCK_SERVER_URL": "http://localhost:29712"
   },
```
Fuzznimp, line 27: Should we keep the `PROFILE_MOCK_SERVER_URL` env?

Fuzznimp:
Question. Would it be easy to have the calendar component 100% of the width in mobile?

src/pages/account/profile/user-profile/update.tsx
Oluane: I made it, but I had conflicts with your last PR merged, and I assumed it was still `update` since on your branch it hadn't changed, and I undid it. I will change it back then.

src/components/input/input-date-picker.tsx
Oluane: Sure

src/pages/account/profile/user-profile/update.tsx
Fuzznimp: Sorry for that. I didn't wanted to make any changes on the file as you were working on it. Quite tricky to all work on the same parts of the app at the same time :/.

Oluane:
@Fuzznimp  Nope, the thing is, as far as I've seen until now, if we want to change the calendar size, we need to scale every sub-components by hand (the line where is the month name, the one with day names and so on...).

Fuzznimp:
Ok, so a time-consuming process. Let's leave it like this for now then.

Oluane:
> Ok, so a time-consuming process. Let's leave it like this for now then.

@Fuzznimp Ok, I manage to scale a little bit the size without too much work after all. I can make it wider if needed

image
image

Meyclem (approved):

Fuzznimp (approved):

PR #516, opened 2021-04-29
Display Connect.Profile data using OpenAPI client

Fuzznimp:
## Description

This PR aims at using the OpenAPI client to display the Connect.Profile data.
I:
- c/p the OpenAPI config file from `connect-profile`.
- Removed the old `get-profile` API route.
- Made some PR amends from the previous PR.

## Type of change

## Issues

There is an issue with the picture URL used in the example data (throws a 404). We should probably start using fake data instead of real world examples.

## Checklist


Fenntasy:
Task linked: [CU-gf7bvb Display profile with Profile client and mock server](https://app.clickup.com/t/gf7bvb)

src/configs/config-variables.ts
```diff
@@ -108,8 +108,7 @@ function handleEnvVars(): void {
     URI: process.env.CONNECT_MANAGEMENT_URL || "",
     APIKey: process.env.CONNECT_MANAGEMENT_API_KEY || "",
   };
-  configVariables.profileMockServerUrl =
-    process.env.PROFILE_MOCK_SERVER_URL || "";
+  configVariables.connectProfileUrl = process.env.PROFILE_MOCK_SERVER_URL || "";
```
Meyclem, line 111:
```suggestion
  configVariables.connectProfileUrl = process.env.PROFILE_MOCK_SERVER_URL || process.env.PROFILE_URL || "";
```
This way, we could just configure `process.env.PROFILE_URL` in the prod environment and make it work in all cases.
Maybe this shouldn't be changed yet though.

src/pages/api/profile/get-profile.ts
```diff
@@ -53,7 +53,10 @@ const handler: Handler = async (request, response) => {
     const { data: profileUserInfo } = await profileClient
       .getProfile()
```
Meyclem, line 54: I think we should have two separate endpoints for the Connect.Profile data:
- `api/connect-profile/get-profile.ts`
- `api/connect-profile/get-addresses.ts`
And make two calls in the pages GSSP using those so we can more easily deal with errors and have a better separation of concerns.

src/components/pages/profile-overview/profile-overview.tsx
```diff
@@ -43,49 +41,49 @@ const ProfileOverview: React.FC<{
           {!data ? (
             <SkeletonTextLine fontSize={1.6} />
           ) : (
-            `${data.userInfo.profile.name} ${data.userInfo.profile.middle_name} ${data.userInfo.profile.family_name}`
+            `${data.profileUserInfo.name} ${data.profileUserInfo.middle_name} ${data.profileUserInfo.family_name}`
```
Meyclem, line 44:
```suggestion
            `${data.profileUserInfo.name}${data.profileUserInfo.middle_name ? " " + data.profileUserInfo.middle_name :  ""} ${data.profileUserInfo.family_name}`
```
Because for now we always have a middlename but it may not always be the case.

Meyclem (dismissed):

src/components/pages/profile-overview/profile-overview.tsx
Fuzznimp: Well, I'd say this is the case for each value, no? As the type is not defined yet, I did not bother checking them first. What do you think?

Meyclem: I approved the PR, so do as you want 😁
We will be forced to check this kind of stuff at some point anyway

src/pages/api/profile/get-profile.ts
```diff
@@ -53,7 +53,10 @@ const handler: Handler = async (request, response) => {
...
       .catch((error) => {
-        span.setDisclosedAttribute("is profile user info fetched", false);
+        span.setDisclosedAttribute(
+          "is Connect.Profile user info fetched",
+          false,
+        );
```
Oluane, line 59: Maybe we could replace `profile` by `Connect.Profile` in this file's other spans, no ?
cf line 49, 76, 81 and 94

Oluane (approved):

PR #526, opened 2021-04-30
First implementation of the address update form

Fuzznimp:
## Description

This PR aims at implementing a naïve update address form without validation. Validation/suggestion will be discussed during the next sprint planning refinement.

- Added a folder in `forms/` as it's getting crowded there. Will need to add a `identity` folder too.
- Added a WIP address show page to access the update flow.
- We don't have the update address endpoint on Connect.Profile yet, will update this when it will be done.

## Type of change

## Screenshots

image

image

## Checklist


Fenntasy:
Task linked: [CU-hh98q5 Implement Address update](https://app.clickup.com/t/hh98q5)

src/components/forms/profile/update-user-address-form.tsx
```diff
@@ -0,0 +1,136 @@
+import React from "react";
+import { v4 as uuidv4 } from "uuid";
+
+import { InputText } from "../../input/input-text";
+// import { WrongInputError } from "../input/wrong-input-error";
```
Meyclem, line 5:
```suggestion
```

src/components/forms/profile/update-user-address-form.tsx
```diff
@@ -0,0 +1,136 @@
...
+    if (userAddress) {
+      setUpdatedAddress(userAddress);
+    }
+  }, [userAddress]);
+
+  return (
+    <>
+      <Form
+        formID={formID}
+        onSubmit={async () => {
```
Meyclem, line 42:
```suggestion
        onSubmit={() => {
```

src/components/forms/profile/update-user-address-form.tsx
```diff
@@ -0,0 +1,136 @@
...
+          name="street-address"
+          placeholder="Enter your street address"
+          value={updatedAddress.street_address}
+          onChange={(value) => {
+            setUpdatedAddress({
+              ...updatedAddress,
+              street_address: value,
+            });
+          }}
+          label="Street address *"
```
Meyclem, line 58: is `*` used for "required"? 🤣

Meyclem: If yes, could we make something like that?
```typescript
type InputTextProps = {
  /** the other props */
  required?: boolean
};

const InputText: React.FC<InputTextProps> = (props) => {
  const { label, className, children } = props;

  const inputRef = React.useRef(null);
  const { labelProps, inputProps } = useTextField(props, inputRef);

  return (
    <Label {...labelProps}>
      {label + required ? " *" : ""}
      {children ? (
        <ChildrenWrapper>
          {children}
          <InputElement {...inputProps} ref={inputRef} className={className} />
        </ChildrenWrapper>
      ) : (
        <InputElement {...inputProps} ref={inputRef} className={className} />
      )}
    </Label>
  );
};
```

src/components/forms/profile/update-user-profile-form.tsx
```diff
@@ -1,9 +1,9 @@
 import React from "react";
 import { v4 as uuidv4 } from "uuid";
 
-import { InputText } from "../input/input-text";
+import { InputText } from "../../input/input-text";
 // import { WrongInputError } from "../input/wrong-input-error";
```
Meyclem, line 5:
```suggestion
```

src/pages/account/profile/addresses/[id]/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { logger } from "@src/configs/logger";
+import getTracer from "@src/configs/tracer";
+import { authMiddleware } from "@src/middlewares/auth-middleware";
+import { sentryMiddleware } from "@src/middlewares/sentry-middleware";
+
+const AddressOverviewPage: React.FC<{ addressId: string }> = ({
+  addressId,
+}) => {
+  const { error } = useSWR<{ address: Address }, Error>(
+    `/api/profile/get-address?addressId=${addressId}`,
```
Comment (Meyclem) [inline src/pages/account/profile/addresses/[id]/index.tsx:28]: In think we should treat this kind of routes as REST endpoints, so instead of that, we should have:
```suggestion
    `/api/profile/addresses/${addressId}`,
```
[1/3]

src/pages/api/profile/get-address.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    if (!userCookie) {
+      response.statusCode = HttpStatus.TEMPORARY_REDIRECT;
+      response.setHeader("location", "/");
+      response.end();
+      return;
+    }
+
+    const addressId = request.query.addressId;
```
Meyclem, line 47: [2/3]
This should be a url param instead of a query param.
The way you wrote this file atm suggest that we would have a `api/profile/update-address.ts` at some point (and maybe a delete and get-addresses, etc...) but I really think we should handle those as REST endpoints:
```
GET /api/profile/addresses
GET /api/profile/addresses/:id
POST /api/profile/addresses
PATCH /api/profile/addresses/:id
DELETE /api/profile/addresses/:id
```

src/pages/api/profile/get-address.ts
```diff
@@ -0,0 +1,115 @@
...
+      throw webErrorFactory(webErrors.invalidQueryString);
+    }
+
+    const profileAccessToken = await getProfileAccessToken(
+      userCookie.access_token,
+    );
+    span.setDisclosedAttribute(
+      "is Connect.Profile access token available",
+      true,
+    );
```
Meyclem, line 59:
```suggestion
    span.setDisclosedAttribute(
      "Connect.Profile access token",
      profileAccessToken,
    );
```
So if there's any error after this line, we can get the token from tracers and use it to make some tests

src/components/forms/profile/update-user-address-form.tsx
```diff
@@ -0,0 +1,136 @@
...
+            setUpdatedAddress({
+              ...updatedAddress,
+              country: value,
+            });
+          }}
+          label="Country *"
+        />
+        <Button type="submit" variant={ButtonVariant.PRIMARY}>
+          Update my address
+        </Button>
```
Meyclem, line 127: What do you think of already making the call to `PATCH /api/profile/addresses/:id` in this task?
Or is it only the front-end implementation?

src/pages/account/profile/addresses/[id]/update.tsx
```diff
@@ -0,0 +1,83 @@
+import {
```
Comment (Meyclem) [inline src/pages/account/profile/addresses/[id]/update.tsx:1]: This page should be called `src/pages/account/profile/addresses/[id]/edit.tsx ` instead of update, because at this point, it's not the update yet, it's only the edit part of the update flow

src/pages/api/profile/get-address.ts
```diff
@@ -0,0 +1,115 @@
...
+    loggingMiddleware(getTracer(), logger),
+    authMiddleware(getTracer()),
+  ],
+  handler,
+  "/api/profile/get-address",
+);
+
+export default new Endpoint<NextApiRequest, NextApiResponse>()
+  .get(wrappedHandler)
+  .getHandler();
```
Meyclem, line 115: [3/3]
An since we already use the `Endpoint` class from fwl/web to deal with that, it should then be:
```suggestion
export default new Endpoint<NextApiRequest, NextApiResponse>()
  .get(wrappedHandler)
  .patch(/** the future patch handler /*)
  .getHandler();
```

Meyclem: (And I think you see me coming for the `identity`'s api pages 😁 )

src/components/forms/profile/update-user-address-form.tsx
Fuzznimp: Damn I forgot to push my last commit t_t

src/components/forms/profile/update-user-address-form.tsx
Fuzznimp: Well, we could, but like I said, I made a task to discuss validation and stuff. The `*` part was just following the figma mock.

Meyclem: 👍

src/pages/api/profile/get-address.ts
Oluane: Isn't access token sensitive data ? 
Shouldn't it be `span.setAttribute` if we decide to put it in the span ?

Meyclem: I don't think so, what we have inside is not private, juste the sub and scopes and the signed part is "only" there to prove it's a valid token

src/components/forms/profile/update-user-address-form.tsx
Fuzznimp: Well, this one is an "issue" with the `Form` component https://github.com/fewlinesco/connect-account/blob/main/src/components/forms/form.tsx. We didn't want to handle sync and async callbacks, so we typed it as async, hence the useless async keyword. I don't really see this as an issue, but if you do, let me know, and we'll make a task to handle both.

src/pages/api/profile/get-address.ts
Fuzznimp: I used to do it this way, but we agreed with fenn many months ago to make an API handler for each HTTP verbs, instead of making one which check the verb asked.

src/components/forms/profile/update-user-address-form.tsx
Fuzznimp: This is only the frontend  implementation :).

Meyclem: Yup...
I just realised the task name was "CU-hh98q5-address_**form_page**" 😅

src/components/forms/profile/update-user-address-form.tsx
Fuzznimp: Yeah, the overhead is negligible, but the "wrongly interpreted" is what bothers me in this, that's why I would be ok to make the changes in the Form component at some points.

src/components/forms/profile/update-user-address-form.tsx
Fuzznimp: Haha no worries, this is the place to ask *any* questions, or remarks :).

src/pages/api/profile/get-address.ts
Meyclem: oh 🤔
@Fenntasy can we have your opinion on this 🙏 ?

src/pages/api/profile/get-address.ts
Fuzznimp: I just saw this comment. Going back to a REST like handling in our file maybe be better 🤔 .

src/pages/account/profile/addresses/[id]/update.tsx
Comment (Meyclem) [reply src/pages/account/profile/addresses/[id]/update.tsx:1]: Still following the RESTful routes, the update form is on the `resources/:id/edit` page.
On this page, the user doesn't really updates its profile, s·he just "edit" the data before the form makes the call to update it.
"It may be a detail for you, but for me it's veut dire beaucoup" 🤣

src/pages/api/profile/get-address.ts
Fuzznimp: Linking this comment as the context is not the same that a yeah ago https://github.com/fewlinesco/connect-account/pull/526#discussion_r623755429

src/pages/api/profile/get-address.ts
Meyclem: Yeah, that's my point, file-wise it would mean a cleaner folder structure (and other things, of course, but it's a part of it)

src/pages/account/profile/addresses/[id]/update.tsx
Comment (Fuzznimp) [reply src/pages/account/profile/addresses/[id]/update.tsx:1]: Well, I see your point, but isn't this just bringing unnecessary confusion between `edit` and `update`?

Comment (Meyclem) [reply src/pages/account/profile/addresses/[id]/update.tsx:1]: As for above, I think @Fenntasy can give us its advice.
Coming from Rails I still have the habit of this naming convention for app which aren't SPA and may lack of objectivity on this

Comment (Fenntasy) [reply src/pages/account/profile/addresses/[id]/update.tsx:1]: > who dares summon me?
> - a confused demon

> bringing unnecessary confusion between edit and update

just because I'm confused there, do we have both at the moment? 

Other than that, in my mind, the web page would be `edit` and the api page would be `update`. And I think of it from the point of view of the user:
- I want to edit my personal data (the web page)
- I want to send my data to something that updates my personal data.

For me the edit is the writing part. 

Does it answer your questions or bring more? 😅

Comment (Fuzznimp) [reply src/pages/account/profile/addresses/[id]/update.tsx:1]: I understand the difference of context between `edit` and `update`. My point was,  is it a necessary to make the distinction here. But I'll change it.

src/pages/api/profile/get-address.ts
Fenntasy: This seems cool, no?

Fuzznimp: Yep, I like it.

Meyclem: https://github.com/fewlinesco/connect-account/pull/526#discussion_r623832669 😇

src/pages/account/profile/addresses/[id]/update.tsx
Comment (Fenntasy) [reply src/pages/account/profile/addresses/[id]/update.tsx:1]: > My point was, is it a necessary to make the distinction here. 

Could you elaborate? Do you think it's simpler to have the same naming on both sides or is that needing to have two verbs or something else?

src/pages/api/profile/get-address.ts
Fenntasy: Yup, saw that but if you both agree in this thread, I don't see the remaining question in the other. 🤔

Meyclem: > we agreed with fenn many months ago to make an API handler for each HTTP verbs, instead of making one which check the verb asked.

I assume then taht it's the change of context @Fuzznimp talked about and we should now use ☝️ ?

src/pages/account/profile/addresses/[id]/update.tsx
Comment (Fuzznimp) [reply src/pages/account/profile/addresses/[id]/update.tsx:1]: My point is, even if the user action is to edit his profile, in the end, all he wants it to update his profile. I don't see why we should split the naming for the flow 🤔 .

src/pages/api/profile/get-address.ts
Fenntasy: Yes, the `Endpoint` was designed to answer this question and the fact that reading that inside the `if`s was a bit weird

src/pages/account/profile/addresses/[id]/update.tsx
Comment (Fuzznimp) [reply src/pages/account/profile/addresses/[id]/update.tsx:1]: > "It may be a detail for you, but for me it's veut dire beaucoup"

I'll make the changes then.

src/pages/api/profile/get-address.ts
Fuzznimp: I'll definitely change it then :).

Meyclem: 👍 🙇

Fuzznimp: I made the changes, and created a task to refacto the rest :).

Fuzznimp: I'll keep these comments unresolved as I point to them in the task.

Meyclem (approved):

PR #541, opened 2021-05-05
Update UserProfile

Fuzznimp:
## Description

This PR aims at adding the base for `UserProfile` update.

Validation are not done in this one, as they need to be discussed Friday during the sprint planning refinement. 

I tried to add the `disabled` attribute set to `true` when the state is loading (to prevent the user starting to input while the data is loading), but the HTML attribute is not set by `react-aria`. I will make a task regarding this.

I also removed the state check inside the date picker component, as we want to handle those on a higher DOM level. There is an issue with the `z-index` of the date picker on chrome, a task is being made to fix it.

## Type of change

## Checklist


Fenntasy:
Task linked: [CU-gf7cea Update UserProfile with OpenAPI client and mock server](https://app.clickup.com/t/gf7cea)

src/components/input/input-date-picker.tsx
```diff
@@ -19,15 +17,11 @@ const InputDatePicker: React.FC<InputDatePickerProps> = ({
...
-        <DatePicker
-          selected={new Date(selected)}
-          onChange={onChange}
-          dateFormat={"yyyy-MM-dd"}
-        />
-      ) : (
-        <SkeletonTextLine fontSize={1.4} />
-      )}
+      <DatePicker
+        selected={selected ? new Date(selected) : undefined}
```
Meyclem, line 21: What is the default when the date is undefined?

If it's not already "today", maybe we could:
```suggestion
        selected={selected ? new Date(selected) : new Date()}
```

src/pages/api/profile/user-profile/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  const webErrors = {
+    invalidProfileToken: ERRORS_DATA.INVALID_PROFILE_TOKEN,
+    invalidScopes: ERRORS_DATA.INVALID_SCOPES,
+    userProfileNotFound: ERRORS_DATA.USER_PROFILE_NOT_FOUND,
+    invalidUserProfilePayload: ERRORS_DATA.INVALID_USER_PROFILE_PAYLOAD,
+    unreachable: ERRORS_DATA.UNREACHABLE,
+  };
+
+  return getTracer().span(
+    "/api/profile/user-profile patchHandler",
```
Meyclem, line 33:
```suggestion
    "PATCH /api/profile/user-profile handler",
```
Maybe more inline with the REST version of writing it?
Wdyt?

src/pages/api/profile/user-profile/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+          throw webErrorFactory(webErrors.unreachable);
+        });
+      span.setDisclosedAttribute(
+        "is Connect.Profile UserProfile updated",
+        true,
+      );
+
+      response.statusCode = HttpStatus.OK;
+      response.json({ updatedUserProfile });
```
Meyclem, line 92: Do we want the api to respond with:
```
{
  updatedUserProfile: {
    name: ...,
    otherFields,
  }
}
```
or directly the resource:
```
{
  name: ...,
  otherFields,
}
```

I think we should use the second version and send the profile rather than an object with a `updatedUserProfile` key, so here's my suggestion:
```suggestion
      response.json(updatedUserProfile);
```

src/@types/http-verbs.ts
```diff
@@ -2,6 +2,7 @@ enum HttpVerbs {
   GET = "GET",
   POST = "POST",
   DELETE = "DELETE",
+  PATCH = "PATCH",
```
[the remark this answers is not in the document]
Fuzznimp: Well, this is like a year old, when I wanted to put enums everywhere 😅 . No reasons to keep it, I'll make the changes 👍 .

src/components/forms/profile/update-user-profile-form.tsx
```diff
@@ -1,31 +1,46 @@
...
+    nickname: "",
+    preferred_username: "",
+    profile: "",
+    picture: "",
+    website: "",
+    gender: "",
+    birthdate: "",
+    zoneinfo: "",
+    locale: "",
+  });
```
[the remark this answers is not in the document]
Fuzznimp: You are right, I just no brained it like I did with addresses 😅 .

src/@types/http-verbs.ts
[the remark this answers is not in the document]
Meyclem: > I wanted to put enums everywhere 😅

🤣

src/components/input/input-date-picker.tsx
Fuzznimp: Well, at first I tried to go with `Date.now()`, which is an invalid type. I forgot that `new Date()` was initializing the constructor with a current date. I'll make the change.

src/pages/api/profile/user-profile/index.ts
Fuzznimp: Yeah, that's what I was proposing on Pauline's PR. Just didn't put any thoughts on this haha. I'll make the changes.

Fuzznimp: Not sure if I should do this now, as it would make this tedious, as each values can be undefined:
```ts
            if ("updatedUserProfile" in parsedResponse) {
              router && router.push("/account/profile");
              return;
            }
            ```

File: src/components/forms/profile/update-user-profile-form.tsx [hunk 6]
```diff
@@ -34,36 +34,96 @@ const UpdateUserProfileForm: React.FC<{ userProfileData?: Profile }> = ({
...
+          }).then(async (response) => {
+            const parsedResponse = await response.json();
+
+            if ("updatedUserProfile" in parsedResponse) {
+              router && router.push("/account/profile");
+              return;
+            }
+
+            throw new Error("Something went wrong");
+          });
```
Meyclem, line 48: Answering to this comment https://github.com/fewlinesco/connect-account/pull/541#discussion_r626550345
I think we can do:

```suggestion
          await fetchJson("/api/profile/user-profile", "PATCH", {
            userProfilePayload: userProfile,
          }).then(async (response) => {
            const parsedResponse = await response.json();
            router && router.push("/account/profile");
            return;
          }).catch(() => throw new Error("Something went wrong");)
```
> Not sure that works though, so maybe more:

```suggestion
          await fetchJson("/api/profile/user-profile", "PATCH", {
            userProfilePayload: userProfile,
          }).then(async (response) => {
            if (response.status !== HttpStatus.OK) {
              throw new Error("Something went wrong");
             }
            const parsedResponse = await response.json();
            router && router.push("/account/profile");
            return;
            }
          });
```

Because the payload can be full of empty fields but should not a problem since the API page will return a "not" 200 code if it's the case. Am I right?

src/pages/api/profile/user-profile/index.ts
Meyclem: Answer here: https://github.com/fewlinesco/connect-account/pull/541#discussion_r626555100

src/components/forms/profile/update-user-profile-form.tsx
Meyclem: Or a mix of the two 🤔 
I mean, If something went wrong on the API side, it will not return the data we want, in any case. So the api should sent a not 200 status code and then, if it's the case, the front should throw.

Fuzznimp: In case of an error, the parsed response will have a `message` key, coming from the web-error, but no `updatedUserProfile `. See https://github.com/fewlinesco/connect-account/blob/main/src/components/forms/update-identity-form.tsx for example.

The first one would not work. The second one would work though.

Also, I can't see any cases for now that we will do anything with the data returned by Connect.Profile in this flow, as we redirect. I'd prefere keeping the same implementation across our code base.

If we wanna change this, we'll have to do it everywhere :).

Wdyt?

Meyclem: but...
> Not sure if I should do this now, as it would make this tedious

Yeah, as you prefer 🙂

Oluane (approved):

src/components/forms/profile/update-user-profile-form.tsx
Meyclem: > Wdyt?

I agree :)

PR #606, opened 2021-05-26
Splitting `get-user-profile-and-addresses` in two separated handlers

Oluane:
## Description

This PR splits the `get-user-profile-and-addresses` handler in two separate handlers, one for `user-profile` and the other for `addresses`. 

## Type of change

- [x] **Breaking change** (fix or feature that would cause existing functionality to not work as expected).

## Checklist


src/components/pages/profile-overview/profile-overview.tsx
```diff
@@ -19,11 +19,9 @@ import {
 } from "@src/utils/format";
 
 const ProfileOverview: React.FC<{
-  data?: {
-    userProfile: Profile;
-    userAddresses: Address[];
-  };
-}> = ({ data }) => {
+  userProfile?: Profile;
+  userAddresses?: Address[];
```
Meyclem, line 23:
```suggestion
  userProfile?: Profile;
  userAddresses: Address[];
```

`Profile` can be undefined if the user doesn't have any, but `userAddresses` shouldn't be because when we send the data from the [API page](https://github.com/fewlinesco/connect-account/blob/d1cf590c181456b51d4b00e5ae418b9f656329bf/src/pages/api/profile/addresses/index.ts#L85) (which comes from Connect.Profile), if there's no address, it's an empty array

src/components/pages/profile-overview/profile-overview.tsx
```diff
@@ -101,27 +99,27 @@ const ProfileOverview: React.FC<{
...
+            quantity={userAddresses.length - 1}
             setHideList={setHideAddressList}
           />
         </Flex>
       ) : null}
-      {data ? (
+      {userAddresses ? (
         <NeutralLink href="/account/profile/addresses/new">
           <FakeButton variant={ButtonVariant.SECONDARY}>
             + Add new address
```
Meyclem, line 125: The button should be always visible, the user should be able to add an address anyway

Meyclem: Or we consider that to have any address, the user should have a mandatory `Profile`?
This kinda makes sense don't you think? Because, what if the purpose to have an address if it is not linked to a name? 🧐

src/pages/account/profile/index.tsx
```diff
@@ -27,27 +27,36 @@ import { sentryMiddleware } from "@src/middlewares/sentry-middleware";
...
-  >(`/api/profile/get-user-profile-and-addresses`);
+  >(`/api/profile/user-profile`);
 
-  if (error) {
-    if (error.statusCode === HttpStatus.NOT_FOUND) {
+  if (userProfileError) {
+    if (userProfileError.statusCode === HttpStatus.NOT_FOUND) {
       router && router.push("/account/profile/user-profile/new");
     }
   }
```
Meyclem, line 39: I don't think we should redirect to the page automatically and replace it with a button to create a Profile maybe?
Wdyt?

src/pages/api/profile/addresses/index.ts
```diff
@@ -21,6 +22,71 @@ import { authMiddleware } from "@src/middlewares/auth-middleware";
...
+      }
+
+      const { addressAccessToken } = await getProfileAndAddressAccessTokens(
+        userCookie.access_token,
+      );
+
+      span.setDisclosedAttribute(
+        "is Connect.Profile access token available",
+        true,
+      );
```
Meyclem, line 56: I just think about that now but we could remove this code from here and put it in the `initProfileClient` client which could be change to `(span: Span) => { addressClient:  ConnectProfileClient, profileClient:  ConnectProfileClient }` or smth like that (but maybe not in this PR, wdyt?)

src/pages/account/profile/index.tsx
```diff
@@ -27,27 +27,36 @@ import { sentryMiddleware } from "@src/middlewares/sentry-middleware";
...
 
+  const { data: userAddress, error: userAddressError } = useSWR<
```
Meyclem, line 41:
```suggestion
  const { data: userAddresses, error: userAddressError } = useSWR<
```

src/pages/account/profile/index.tsx
```diff
@@ -27,27 +27,36 @@ import { sentryMiddleware } from "@src/middlewares/sentry-middleware";
...
   return (
     <Layout
       breadcrumbs="Basic information about you"
       title="Personal information"
     >
       <Container>
-        <ProfileOverview data={data} />
+        <ProfileOverview
+          userProfile={userProfile}
+          userAddresses={userAddress}
```
Meyclem, line 58:
```suggestion
          userAddresses={userAddresses}
```

Meyclem (changes requested):

src/components/pages/profile-overview/profile-overview.tsx
Oluane: I asked myself this question yesterday, and I finally decided to display it conditionally to the moment `userAddress` was defined. So until data is ready to be displayed, we have skeleton component for addresses, and I wanted to hide the button during this time, and display it with the data. It's more a "display/UX" concern in fact.

src/components/pages/profile-overview/profile-overview.tsx
Oluane: I think the possibly undefined here is more about SWR data fetching way than Profile possible response. I'm gonna test it to see.

src/components/pages/profile-overview/profile-overview.tsx
Oluane: We did the same for `identities` buttons if I remember well, just to be sure that a user won't try to click buttons before data arrived client side

src/pages/account/profile/index.tsx
Oluane: That's a real question indeed. We said we would do that way in the first place, but that could be great to discuss that again with the team. Personally, I kinda like that way, because it feels "automatic", less click to perform an action, but I'm not close to add a button instead 🙂

src/components/pages/profile-overview/profile-overview.tsx
Oluane: Yep, I confirm. The possibly `undefined` here is just here because we fetch data client side with SWR. 

image

src/pages/api/profile/addresses/index.ts
Oluane: You mean, with all cookies related things ? 🤔

src/components/pages/profile-overview/profile-overview.tsx
Meyclem: > the moment userAddress was defined

That's my point, I think it should always be defined, at least as `[]`

But for 

> the user should have a mandatory Profile?

Addresses or not, if s·he don't have a profile, s·he shouldn't be able to create and address, so what I think is that we should change `userAddresses ?` to `userProfile ?` for the `create an address` button

src/pages/account/profile/index.tsx
Meyclem: I like it both ways and as you said, we should discuss it 🙂
(Can you create a task about that?)

src/pages/api/profile/addresses/index.ts
Meyclem: Yep, it is redundant and we repeat exactly the same code in each Profile related endpoint

src/components/pages/profile-overview/profile-overview.tsx
Meyclem: Yes, but when passing the data to the component here: https://github.com/fewlinesco/connect-account/blob/8a8fff5c6bc33f2c0a9fc783b05643c60ba3ef4b/src/pages/account/profile/index.tsx#L58

Maybe we could have a default value to `[]` in case of undefined here: https://github.com/fewlinesco/connect-account/blob/8a8fff5c6bc33f2c0a9fc783b05643c60ba3ef4b/src/components/pages/profile-overview/profile-overview.tsx#L23

Oluane: Yep. And how should we do for the Skeleton component if we have a default value equal to `[]` then ? I mean, we couldn't rely on the fact that `userAddresses` is an empty array, because it could mean in this case that either data isn't ready yet or that there isn't any data at all.

src/components/pages/profile-overview/profile-overview.tsx
Oluane: > Addresses or not, if s·he don't have a profile, s·he shouldn't be able to create and address, so what I think is that we should change userAddresses ? to userProfile ? for the create an address button

Indeed. But as it is done for now, this isn't an issue because you get redirected to the creation of profile form if you haven't any profile yet, so you never have the chance to see this page and this button 🙂 

But, if we decide to go for a "create profile" button, we should then change that, I agree.

src/pages/account/profile/index.tsx
Oluane: Yep, I will create it. 👍

src/components/pages/profile-overview/profile-overview.tsx
Meyclem: Actually, if we don't have any address, that means we:
- display de skeleton component until we get the response
- remove it (because no address)
- display a button

And it feels weird too don't you think?

src/pages/api/profile/addresses/index.ts
Meyclem: > but it should definitely be done in another PR if we go for that

I agree 😁

src/components/pages/profile-overview/profile-overview.tsx
Fenntasy: I feel this is just a naming problem. This component receives data from a SWR call (from what I understand) which has two values for the data received: `undefined` and what we type it with. 

Which means that we don't pass to this component the data really but a "state of the data" (or a representation of the data). I feel there's two way to reconcile your two points of view, either renaming `userAddresses` into something like `userAddressesData` and agreeing that it's not yet the real data (which could only be an array) or else have the component receive an object that could be typed something like this:

```ts
type UserProfileData = "loading" | { userProfile: Profile; userAddresses: Address[] }
```

(the `"loading"` part could be an object with something to mark that it's in a loading state)

That way, it's only when the component received a loading state that we display the skeletton, not when we have the data (whateter form it has).

What do you think?

Meyclem: > What do you think?

👇 

> (the "loading" part could be an object with something to mark that it's in a loading state)

I totally agree with that

> renaming userAddresses into something like userAddressesData and agreeing that it's not yet the real

Less with that part because I don't think the address component should treat with undefined data because this specific data can't be (not like the Profile one)

Meyclem: But if we go on the `loading` road, this should be done in this PR though

Oluane: >(the "loading" part could be an object with something to mark that it's in a loading state)

I really like this one too. 

I guess I'm gonna do a task for that then 🙂

src/pages/account/profile/index.tsx
```diff
@@ -27,27 +27,36 @@ import { sentryMiddleware } from "@src/middlewares/sentry-middleware";
...
+    Profile,
     SWRError
-  >(`/api/profile/get-user-profile-and-addresses`);
+  >(`/api/profile/user-profile`);
 
-  if (error) {
-    if (error.statusCode === HttpStatus.NOT_FOUND) {
+  if (userProfileError) {
+    if (userProfileError.statusCode === HttpStatus.NOT_FOUND) {
       router && router.push("/account/profile/user-profile/new");
```
Meyclem, line 37:
```suggestion
      router && router.replace("/account/profile/user-profile/new");
```
Because otherwise it is added to the history and if the user goes back after the redirect it is redirect again

Oluane: Very good point 👍

src/components/pages/profile-overview/profile-overview.tsx
Fenntasy: Here's an exemple of what I had in mind https://www.typescriptlang.org/play?#code/C4TwDgpgBACgTgewGYEsA20C8UDeVgIEBcUAjFAL4BQVokUAqgM4RzzLoQAiAhsD1Gx4m-YBBIAiNAh4ATFADsA5hMpQAPrigi+4qFJmyIsiQBoosvjxLtUGSgG4aAYwQKRUAK4s2iO9ysSZlZbTl5+QSgAWT4ACwA6OB4FWQQAWwAKAEooAB4oAAZ4gFYoAH4tHTFJaTlFFQooEhwqvQM5YzNLfmb8QgQSUgpqKld3YHwID2wM7xC-MKt41sFMbHb5ZQkcqig9-b2KhU80NChdg-2SOd8ODHCeeO7HggIqIA

Meyclem (approved):

PR #614, opened 2021-05-26
Delete address

Oluane:
## Description

This PR adds the delete address flow. 

## Type of change

## Checklist


Fenntasy:
Task linked: [CU-k35ghv [feature] - Delete an address](https://app.clickup.com/t/k35ghv)

src/pages/api/profile/addresses/[id]/index.ts
```diff
@@ -175,6 +174,73 @@ const patchHandler: Handler = async (request, response) => {
...
+        if (error.response.status === HttpStatus.NOT_FOUND) {
+          throw webErrorFactory(webErrors.addressNotFound);
+        }
+
+        throw webErrorFactory(webErrors.unreachable);
+      });
+
+      span.setDisclosedAttribute("is Connect.Profile address deleted", true);
+
+      response.statusCode = HttpStatus.ACCEPTED;
```
Comment (Meyclem) [inline src/pages/api/profile/addresses/[id]/index.ts:238]:
```suggestion
      response.statusCode = HttpStatus.NO_CONTENT;
```

Meyclem (changes requested):

src/pages/api/profile/addresses/[id]/index.ts
Comment (Oluane) [reply src/pages/api/profile/addresses/[id]/index.ts:238]: Hmmm could you explain me the "why" of this change ? Isn't `202` status code appropriated for this flow? 🤔 (I ask that because if it isn't, we should have another place to change, in the delete identity flow)

Comment (Meyclem) [reply src/pages/api/profile/addresses/[id]/index.ts:238]: Yep: https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.7

So:
- `200` is ok in each case, but tbh, I don't like it because it's just "ok" and you don't know exactly what happened
- `202` is "accepted" too but **"the action has not yet been enacted"** meaning, the DELETE part is planed but not done yet.
- `204` is "accepted" and done, not in the `202` case and says clearly that we don't send any body with the response which is the case for the Connect.Profile API.

So here, we should change it to `NO_CONTENT` but for the entities, I don't know if Management does the deletion immediately or not.
- If not, the `202` is correct
- if yes:
  - if something is returned in the response, `200`
  - if nothing is returned in the response, `204`

More broadly, and this is just for the communication with Connect.Profile, I think the api pages of Account should always respond with the same body and code that Profile does.

Comment (Oluane) [reply src/pages/api/profile/addresses/[id]/index.ts:238]: Yes, I get the distinction and it's clearer. And now that you said that, I remember that we put `202` on identities deletion flow because indeed, Management takes into account that we want to delete an identity, but doesn't do it immediately. 

>More broadly, and this is just for the communication with Connect.Profile, I think the api pages of Account should always respond with the same body and code that Profile does.

Yep, I agree with that. I'm gonna make the change 🙂

PR #654, opened 2021-06-03
Profile e2e tests

Oluane:
## Description

This PR adds an e2e tests for Profile and Address happy path. The both parts of Profile are tested in the same test file, and are designed to work for our two test environment, review apps with mock servers and staging with Profile staging with conditional statement on the Profile URL set in the env vars. 

I modified the step that is happening just before the e2e test runs, which creates the test user on Connect side. I get back the user `sub` to be able to do cleaning on Profile staging with the admin endpoints once the tests are ran. 

For the cleaning part, we just have to delete the user profile. For the addresses, there was an issue with getting the created during tests addresses ids, so I decided to add a conditional-to-staging-env step in the e2e tests to delete the address remaining after the test flow directly with Taiko. The cleaning user profile step will only be run in CI if the Account environment is staging. 

For the CI part, I added two conditional steps in our `config.yml` for CircleCI (cf above ⬆️ ) regarding staging env. The other one is to set the Profile URL env var dynamically. 

## Type of change

## Issues

I had big issues with Taiko's speed while writing this new and quite long test flow. After a lot of trials, I found out that enabling `observe` mode browser option ([doc](https://docs.taiko.dev/api/openbrowser/)) with an `observeTime` value between each action perfomed set to 500ms for this test flow was the best way to go if we want to have a more reliable tests suite, less "random" error-prone. 
It surely makes our test time greater (the `profile-happy-path.test.ts` lasts for more or less a 1 minute by itself), but imho, it's better than having to rerun tests several time because some randomly and unpredictable fails occurs (because of what I suppose to be Taiko's speed vs the application speed).

In the same topic, I also make the call to not rely on alert messages appearence to check if a flow has succeeded, but rather relied on checking the current url. I was having troubles with checking alert messages, their appearence seems to have some inconsistency in Taiko's browser environment (maybe again a matter of speed/time).

## Screenshots 

image
Screenshot of tests runned on review app against Profile staging

## Checklist


frieda-ewlines:
Task linked: [CU-kvcv6f [test] - Add `Connect.Profile` happy path](https://app.clickup.com/t/kvcv6f)

tests/e2e/profile-happy-path.test.ts
```diff
@@ -0,0 +1,148 @@
...
+      );
+
+      await click("Add address");
+      await waitFor(200);
+      expect(await currentURL()).toEqual(
+        `${
+          process.env.CONNECT_TEST_ACCOUNT_URL ||
+          configVariables.connectAccountURL
+        }/account/profile`,
+      );
```
Oluane, line 91: I had to replace the alert message existence check by the check of the current URL here, because Taiko wasn't able to find the alert message (once again, this is an issue with Taiko speed while not on headless mode, because when you run the test file with headless mode, it works as expected.) I didn't find any other working workaround, and since when the flow (here adding an address) is successful, we are redirected to a precise URL, I thought it would count as a valid check.

tests/e2e/profile-happy-path.test.ts
```diff
@@ -0,0 +1,148 @@
...
+        await text("You are about to delete this address.").exists(),
+      ).toBeTruthy();
+      await click("Delete");
+      await waitFor(200);
+      expect(await currentURL()).toEqual(
+        `${
+          process.env.CONNECT_TEST_ACCOUNT_URL ||
+          configVariables.connectAccountURL
+        }/account/profile`,
+      );
```
Oluane, line 121: Same as above ☝️

tests/e2e/profile-happy-path.test.ts
```diff
@@ -0,0 +1,143 @@
...
+        into(textBox({ placeholder: "Enter your postal code" })),
+      );
+      await write(
+        "United States of America",
+        into(textBox({ placeholder: "Enter your country" })),
+      );
+
+      await click("Add address");
+      await waitFor(250);
+      expect(await currentURL()).toEqual(`${baseURL}/account/profile`);
```
Oluane, line 91: I had to replace the alert message existence check by the check of the current URL here, because Taiko wasn't able to find the alert message (once again, this is an issue with Taiko speed while not on headless mode, because when you run the test file with headless mode, it works as expected.) I didn't find any other working workaround, and since when the flow (here adding an address) is successful, we are redirected to a precise URL, I thought it would count as a valid check.

tests/e2e/profile-happy-path.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      expect(await text("Your address has been updated").exists()).toBeTruthy();
+
+      // Delete address flow
+      await click("Delete this address");
+      expect(
+        await text("You are about to delete this address.").exists(),
+      ).toBeTruthy();
+      await click("Delete");
+      await waitFor(250);
+      expect(await currentURL()).toEqual(`${baseURL}/account/profile`);
```
Oluane, line 116: Same as above ☝️

Meyclem:
> This test is designed to be used with Profile mock server only

How will we manage to **not** run this test in staging then?

Oluane:
> How will we manage to not run this test in staging then?
And why / how is it designed only for the mock server?

Don't we tell last week that staging should be plugged to a mock server and that only prod will use a real Profile version? 🤔

Fuzznimp (approved):
Awesome 😮

PR #670, opened 2021-06-08
Implement locale handling

Fuzznimp:
## Description

This PR aims at adding the locale handling through the UI. This is a preparatory work for the incoming i18n feature.

- Changed the navigation label from "English" to "Language". It would be great to have the actual local as label, but this would add more complexity to an already big PR. I will create a task for it.
- Removed the search input as it is unnecessary, and updated the integration tests accordingly.
- Renamed `WrongInputError` to `ErrorMessage`.
- Added an alert message.
- Loading state UI.

I did not make any e2e tests, as i18n is not in place yet, and we are in the process of migrating our CI.

We discussed to maybe do a middleware, but I did not see the use for it. Let me know if you have any suggestions.

## Type of change

## Screenshots

image

## Checklist


frieda-ewlines:
Task linked: [CU-m3dxp5 [feature] - Implement locale handling (middleware?)](https://app.clickup.com/t/m3dxp5)

src/components/pages/locale/locale.tsx
```diff
@@ -1,71 +1,77 @@
...
+              const parsedResponse = await response.json();
+
+              if ("message" in parsedResponse) {
+                setErrorMessage("Something went wrong. Please try again later");
+                setFormID(uuidv4());
+              }
+
+              if ("updatedUser" in parsedResponse) {
+                router && router.push("/account");
+              }
```
Fuzznimp, line 53: @Fenntasy Would you rather use this or the response's HTTP status code?

src/pages/api/locale/get-locale.ts
```diff
@@ -0,0 +1,81 @@
+import { Endpoint, getServerSideCookies, HttpStatus } from "@fwl/web";
```
Oluane, line 1: You should rename this file `index.ts` and declare the two locale handlers here, to follow our other endpoints. 🙂

Fuzznimp: Good point. 😮

Fuzznimp:
I forgot to add the alert message :/.

src/components/pages/locale/locale.tsx
Fuzznimp: Also, while working on adding an alert message, I just realized that do not return the `Item`, but this
```
{
connect-account |   '$metadata': {
connect-account |     httpStatusCode: 200,
connect-account |     requestId: '90b46643-081d-40f1-b144-6be77ca49c7c',
connect-account |     extendedRequestId: undefined,
connect-account |     cfId: undefined,
connect-account |     attempts: 1,
connect-account |     totalRetryDelay: 0
connect-account |   },
connect-account |   Attributes: undefined,
connect-account |   ConsumedCapacity: undefined,
connect-account |   ItemCollectionMetrics: undefined
connect-account | }
```

src/pages/api/locale/index.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    span.setDisclosedAttribute("user locale retrieved", true);
+    span.setDisclosedAttribute("user locale value", currentDBUser.locale);
+    response.statusCode = HttpStatus.OK;
+    response.json({ locale: currentDBUser.locale });
+    return;
+  });
+};
+
+const postHandler: Handler = (request, response): Promise<void> => {
```
Meyclem, line 77: Wdyt of using `PATCH` instead?
I mean, it's more a `locale` modification, rather than a creation, right?

Oluane (dismissed):

src/pages/api/locale/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      .catch((error) => {
+        span.setDisclosedAttribute("user locale set", false);
+        throw webErrorFactory({
+          ...webErrors.databaseUnreachable,
+          parentError: error,
+        });
+      });
+
+    span.setDisclosedAttribute("user locale set", true);
+    response.statusCode = HttpStatus.OK;
```
Meyclem, line 123: If we use a `POST` method, it should be
```suggestion
    response.statusCode = HttpStatus.CREATED;
```

src/pages/api/locale/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      response.statusCode = HttpStatus.TEMPORARY_REDIRECT;
+      response.setHeader("location", "/");
+      response.end();
+      return;
+    }
+
+    span.setDisclosedAttribute("user locale retrieved", true);
+    span.setDisclosedAttribute("user locale value", currentDBUser.locale);
+    response.statusCode = HttpStatus.OK;
+    response.json({ locale: currentDBUser.locale });
```
Meyclem, line 72: Following REST convention we should only send back the resource and not embed the data in an object with a `locale` key.
```suggestion
    response.json(currentDBUser.locale);
```

Fuzznimp: I did this because using `response.json(currentDBUser.locale);` throws a `JSON SyntaxError` because of the nested accessed value inside the object. I would need to declare a variable beforehand to do so. But I can do it if you want to.

src/pages/api/locale/index.ts
Fuzznimp: I agree !

src/pages/api/locale/index.ts
Meyclem: What do you mean?
Isn't it just a string like "fr" or "en"?

Fuzznimp:
```ts
// Throws
response.json(currentDBUser.locale);

// Works
const updatedLocale = currentDBUser.locale;
response.json(updatedLocale);
```

Meyclem: 😮 What is this wizardry? 🤔
And in both cases, is `currentDBUser.locale` has a value?
If the locale should be "en" by default, maybe this could do the trick?
```ts
response.json(currentDBUser.locale || "en");
```
(sorry, but i'm very curious on that partlicuar magic 😁)

Meyclem: >  also get `SyntaxError: Unexpected token e in JSON at position 0.`. Only returning an object is working 🤯 .

Maybe with `response.end` instead, like we do in the Profile's ping handler?

```ts
export function pingHandler(tracer: Tracer) {
  return (
    request: IncomingMessage,
    response: ServerResponse,
  ): Promise<void> => {
    return tracer.span("ping-handler", async () => {
      response.statusCode = HttpStatus.OK;
      response.end("OK"); 👈👈👈
    });
  };
}
```

Fuzznimp: Not working :/. Would you be ok to fetch the branch and verify I'm not crazy? Haha.
I just pulled the switch from POST to PATCH 👍 .

Meyclem: I'm ok, I'm on it

Oluane (approved):

PR #680, opened 2021-06-10
Setup CircleCI

Oluane:
## Description

This PR is the first step of our migration from GH actions to CircleCI, with the addition of the project configuration and for now, the migration of the `lint` and `tests` jobs. 

## Type of change

## Checklist


frieda-ewlines:
Task linked: [CU-m9c46j [ci] - Migrate build and lint workflows to CircleCI](https://app.clickup.com/t/m9c46j)

.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
...
+  linter:
+    docker:
+      - image: cimg/node:14.17.0
+    steps:
+      - checkout
+      - run:
+          name: lint
+          command: | 
+            yarn install
+            yarn lint
```
Fuzznimp, line 12: We should use caching here, don't you think?

Oluane: Oh you're right 😅

Meyclem: Or you can run two steps inside the same job, as I do on Profile: https://github.com/fewlinesco/connect-profile/blob/04634f2ca90db8a48bf5a2dd9e362ab526e973ec/.circleci/config.yml#L32

Oluane: Don't we want to have them separated?

Meyclem: Well, I think it is ok in both cases and I don't see why we would like to separate them.
I mean, this is part of the testing process and separate them just takes more time to run the CI.

It will not be the case with e2e which would need a separate job for sure, but here, since they use exactly the same thing...

Oluane: But... what is the behaviour if the first one fails ? Is the second executed anyway or the workflow stops there ?

Fuzznimp: Good question.

Meyclem: the jobs stops

Oluane: Maybe @Fenntasy could enlighten us on this ? 🙂

Fenntasy: Well that's more a choice on you folks. 
Having them on the same job means that the first one failing will mask the other one (so if we have errors in tests and a lint problem, they would need to be fixed separately).
Having them on another job means that it's slower (or costlier since I think we could run them concurrently).

However, both have merits.

Fuzznimp (approved):

PR #698, opened 2021-06-16
Cu mf6uaz new endpoint refresh token flow

Oluane:
## Description

This PR refacto the way we do authentication and refresh of tokens when they are expired. The motivation behind this was the discovery of an issue with the refresh flow and the use of Profile api. Indeed, to contact Profiles endpoints, we need to contact Connect first to do a token exchange (given the Account `access_token`, Connect sends us back either the profile and address one) via `/userinfo`. But with the way `authMiddleware` and the refresh token flow used to work, when the Account token was expired, we indeed refreshed the token, but used the expired one to contact Connect, and never get the Profiles tokens. (we take the token from the `request`, and put the refreshed one in the `response`, so once we arrive in the handler, we use the "old" token, with no access to the new one inside the `response`.)

We had multiple solutions to fix this, and we chose to take the occasion to refacto a bit the way it all works, by doing a separation of concerns. Now, `authMiddleware` just does `access_token` validity check, and redirects request to the refresh token flow, which is an endpoint now, if needed. We store the requested page inside a `next` query params, to be able to redirects to it if the refresh flow succeed. 

I tested manually this whole flow several time until the tests case I could think of passed. (waiting an hour between each try and so on 😆 ). 

## Type of change

## Checklist

- [ ] I have performed a self-review of my own code.


frieda-ewlines:
Task linked: [CU-mf6uaz [feature] - Rework of refresh token flow](https://app.clickup.com/t/mf6uaz)

src/middlewares/auth-middleware.ts
```diff
@@ -48,104 +47,24 @@ async function authentication(
...
-          });
-        });
-        span.setDisclosedAttribute("database reachable", true);
-
-        if (!user) {
-          span.setDisclosedAttribute("user found on DB", false);
-          throw new NoDBUserFoundError();
+        if (request.url && request.url.includes("/api/")) {
+          throw webErrorFactory(webErrors.unauthorized);
         }
```
Oluane, line 58: Here, I filter the request to api pages, to avoid having multiple redirection at the end of refresh token flow. I throw an unauthorized web error, which will be catch in `_app`. (cf further comment)

src/pages/_app.tsx
```diff
@@ -42,6 +45,14 @@ const AccountApp: React.FC = ({ children }) => {
...
                   );
 
+                  if (
+                    response.status === HttpStatus.UNAUTHORIZED &&
+                    response.statusText === "Unauthorized"
+                  ) {
+                    router && router.replace(router.pathname);
+                    return;
+                  }
+
```
Oluane, line 55: Here, I catch the unauthorized web error, and trigger a page reload to retrigger all auth flow. This piece is here because of the revalidate on focus feature from SWR, where when you come back on a left opened tab, the api calls are retriggered to update data. But, as the auth middleware ignores the api calls to do refresh flow, we need to reload the page to make everything work smoothly.

src/errors/web-errors.ts
```diff
@@ -14,6 +14,11 @@ type WebErrorData = {
 };
 
 const ERRORS_DATA: Record<string, WebErrorData> = {
+  TEMPORARY_REDIRECT: {
+    code: "temporary_redirect",
+    httpStatus: HttpStatus.TEMPORARY_REDIRECT,
+    message: "Temporary redirection",
+  },
```
Fuzznimp, line 21: Even if `307` are not considered errors, I'm ok with this being here.

src/middlewares/auth-middleware.ts
```diff
@@ -48,104 +47,24 @@ async function authentication(
...
-            ...webErrors.databaseUnreachable,
-            parentError: error,
-          });
-        });
-        span.setDisclosedAttribute("database reachable", true);
-
-        if (!user) {
-          span.setDisclosedAttribute("user found on DB", false);
-          throw new NoDBUserFoundError();
+        if (request.url && request.url.includes("/api/")) {
```
Fenntasy, line 56:
```suggestion
        if (request.url && request.url.startsWith("/api/")) {
```

Seems safer to me, what do you think?

src/middlewares/auth-middleware.ts
```diff
@@ -48,104 +47,24 @@ async function authentication(
...
-        ).catch((error) => {
-          span.setDisclosedAttribute("database reachable", false);
-          throw webErrorFactory({
-            ...webErrors.databaseUnreachable,
-            parentError: error,
-          });
-        });
-
-        span.setDisclosedAttribute("user updated on DB", true);
+        const redirectURL = !request.url ? "/account" : request.url;
```
Fenntasy, line 60: 🙃 Isn't simpler to:?
```suggestion
        const redirectURL = request.url ? request.url : "/account";
```

src/middlewares/auth-middleware.ts
```diff
@@ -48,104 +47,24 @@ async function authentication(
...
 
-        return;
+        response.statusCode = HttpStatus.TEMPORARY_REDIRECT;
+        response.setHeader(
+          "location",
+          `/api/auth-connect/refresh-token?next=${redirectURL}`,
```
Fenntasy, line 65: does it always work? Shouldn't we use `encodeURIComponent`?

src/middlewares/auth-middleware.ts
Oluane: "Why make it simple when you can make it complicated" 😅   
Joke aside, your suggestion is far better and easier to read.

src/middlewares/auth-middleware.ts
Oluane: I feel like a fool, I didn't see your review until this afternoon. (Once again, I didn't receive notifications on Slack for the comments... 😖 )
But regarding the comment, I agree, it will be safer I think 🙂

src/middlewares/auth-middleware.ts
Oluane: It worked when I tested it during development, but always, I don't know. Anyway, I didn't knew `encodeURIComponent` before looking at your comment, and just in case, it seems safer to add it. 

Is it mean to be used like that ? 
`/api/auth-connect/refresh-token?next=${encodeURIComponent(redirectURL)}`

Fenntasy: yes

Fenntasy (approved):

Meyclem (approved):

PR #721, opened 2021-06-22
Improving e2e tests performance and clarity

Oluane:
## Description

This PR was the occasion to go through our e2e tests suite and try to improve either performance or clarity, or both. Here is the main changes : 

* `password-flow.test.ts` : to make this test runnable again, I added a new script, a new npm script and a new step in our `config.yml` to execute these scripts. To test filling input with wrong values in the password flow, we need to "bypass" the sudo mode, so the new script adds an entry for the test user in our DynamoDB table with a valid `sudo_mode_ttl`. This script works in local env too, if needed. I consider adding it to the custom local npm script for e2e (?).

* `mark-identity-as-primary.test.ts` :  there was an issue with this test, where if it was ran before the delete identity one, it used the identity to delete to perform the mark as primary action. I refacto this test for it to never select the to-delete identity, but one of the 2 others identities instead. I cleaned up some useless expect assertions too. 

* `delete-identity.test.ts` : the way we tested that the to-delete identity wasn't present anymore at the end of the test was based on a error prone expect assertion. It is quite difficult to test the absence of an element with tools provided by Taiko (at least, the ones I know about), even more with the delay to update data coming from SWR. But I think I managed to improve it a bit, although it's not perfect yet. The test now checks the first identity link doesn't contain "delete", then the second one and finally, the next link below the second one (which is logically the "+ Add a new address email"). This way, there's no timeout, nor false-positive results. I enabled the `observe` mode on this test too, with an `observeTime` sets to 500ms, to slow down Taiko, and let some time to the app to refresh data. I cleaned up some useless expect assertions too. 

For the other ones, I mainly cleaned up some useless expect assertions. 

Overall, on performance side, the real improvement seems to be on `delete-identity.test.ts`. It lasts now between 15 and 20s, instead of 45-50s before the refacto. I don't know yet if the others are really faster now. 

## Type of change

## Checklist


frieda-ewlines:
Task linked: [CU-mxejhy [refacto] - Improving performance and clarity of e2e tests](https://app.clickup.com/t/mxejhy)

Fuzznimp (commented):
Awesome work!
I agree that you should add the sudo npm script to the local e2e npm script 👍

Fuzznimp (approved):

PR #786, opened 2021-07-07
change profile happy path e2e

Meyclem:
## Description

This PR is a first step in the improvement of our e2e tests flows.
I mostly worked on the "profile and addresses happy path" test and added some tooling.

## Type of change

## Issues

[CU-9u11zz](https://app.clickup.com/t/9u11zz)

## Checklist


frieda-ewlines:
Task linked: [CU-9u11zz Improve the way of running e2e tests in local env](https://app.clickup.com/t/9u11zz)

.env_vars.sample
```diff
@@ -46,5 +46,6 @@ export CONNECT_TEST_ACCOUNT_PASSWORD=
 export CONNECT_TEST_ACCOUNT_USER_ID=
 export CONNECT_ACCOUNT_TEST_URL=
 
-## Connect.Profile
-export CONNECT_PROFILE_URL=
+## Profile
+export CONNECT_PROFILE_URL=http://localhost:21001
+export PROFILE_API_KEY="<api-key-here>"
```
Meyclem, line 51: I added a script to delete the test user from Connect.Profile's database and we need that to be authorized on the targeted endpoint.

bin/e2e/delete-profile-user.ts
```diff
@@ -0,0 +1,29 @@
...
+      console.log(
+        `✅ Profile with sub '${process.env.CONNECT_TEST_ACCOUNT_SUB}' deleted from Connect.Profile`,
+      );
+    }
+  })
+  .catch((error) => {
+    if (!(error.message === "Request failed with status code 404")) {
+      console.error("💥", error.message);
+      process.exit(1);
+    }
```
Meyclem, line 28: The user may not exist, mainly for two reasons:
- we are testing against the mock server
- It has already been deleted
When it is the case, we don't want the script to fail.

package.json
```diff
@@ -14,7 +14,8 @@
     "start": "next start -p $PORT",
     "test": "jest --testPathIgnorePatterns=e2e --runInBand",
     "test:e2e": "jest --testPathPattern=e2e --runInBand",
-    "test:e2e:local": "ts-node bin/e2e/prepare-e2e-tests.ts && ts-node bin/e2e/setting-sudo-ttl.ts && jest --testPathPattern=e2e --runInBand",
+    "test:e2e:local": "yarn test:e2e:delete-profile-user && yarn test:e2e:remove-screenshots && ts-node bin/e2e/prepare-e2e-tests.ts && ts-node bin/e2e/setting-sudo-ttl.ts && jest --testPathPattern=e2e --runInBand",
```
Meyclem, line 17: The `deleteProfileUser` script is started each time we run the e2e tests in local.
It avoids us having to manually do it, either directly in the Profile's database, either with a call to the admin endpoint.

Let me know if you prefer to run it separately.

tests/e2e/utils/authenticate-to-connect.ts
```diff
@@ -13,20 +13,25 @@ async function authenticateToConnect(): Promise<void> {
...
-
-    const needScopeAcceptance = await text(
-      "would like to have access to these information about you",
-    ).exists(10000, 50);
-
-    if (needScopeAcceptance) {
-      await click("Accept");
+    const isAlreadyLoggedIn = await text("already logged in").exists(1000, 50);
+    if (isAlreadyLoggedIn) {
+      await click("Continue");
```
Meyclem, line 18: When the user is already logged in there's the "Continue" button instead of the login form.
It happens on the Profile happy path since I separated profile and addresses tests (in the same file), so I added it in this function.

package.json
Fuzznimp: I'm fine with this 👍.

bin/e2e/delete-profile-user.ts
Oluane: We already had this script (https://github.com/fewlinesco/connect-account/blob/main/bin/e2e/cleaning-profile-staging.ts), which is used in `config.yml` (l70)

tests/e2e/utils/authenticate-to-connect.ts
Fuzznimp: We'll need to remember this when we'll refactor the auth middleware to redirect to `/account` from `/` when the user is already logged in.

Fuzznimp (commented):
Good for me 👍.

bin/e2e/delete-profile-user.ts
Meyclem: Yes, but since it's tailored for staging I preferred to add it only for e2e.
Do you prefer me to tweak this and use the script I wrote in the code you shared to avoid repetition?

Oluane: It isn't really tied to staging, there's just the API_KEY env var to change and it could be used anywhere I think. The script name could also be changed, to not reference staging. Anyway, do as you want, but we don't need to have the both imo.

Meyclem: I changed the code:
- removed my function
- tweaked a bit the existing one to be less specific to staging (mainly naming).

I also changed the env variable name and added it to CircleCi (so it is there 2 time for now with two different names but I'll delete the `CONNECT_PROFILE_STAGING_API_KEY` one just before the merge of this PR).

tests/e2e/profile-happy-path.test.ts
```diff
@@ -45,36 +52,31 @@ describe.only("Profile happy path", () => {
...
   afterAll(async () => {
     await closeBrowser();
   });
 
   const isStagingEnv = (
     process.env.CONNECT_PROFILE_URL || configVariables.connectProfileUrl
   ).includes("staging");
+  // const isStagingEnv = (
+  //   process.env.CONNECT_PROFILE_URL || configVariables.connectProfileUrl
+  // ).includes("staging");
```
Oluane, line 68: Is it intended ?

tests/e2e/profile-happy-path.test.ts
```diff
@@ -85,29 +87,29 @@ describe.only("Profile happy path", () => {
         );
 
         await click(localizedStrings.newProfile.add);
-        expect(await currentURL()).toEqual(profileUrl);
```
Oluane, line 88: Why did you remove this except ? It was meant to verify that the add profile flow worked as expected, since if it is successful, we are redirected to the Profile overview.

tests/e2e/utils/authenticate-to-connect.ts
Oluane: It could be great for consistency to update `fill-incorrect-identity-input-values.test.ts`, which is the only other test file that has two separated test cases, and make it use what you did with `profile-happy-path.test.ts` and `authenticate-to-connect.ts`

Meyclem: Done 👍

tests/e2e/profile-happy-path.test.ts
Meyclem: It is useless in fact, because if we aren't on that page, the line 👇  will fail and will also make the test fail.
```ts
await click(localizedStrings.profileOverview.updateInfo);
```

At first I forgot to put it back but it doesn't matter so I didn't put it back 😁

tests/e2e/profile-happy-path.test.ts
Meyclem: Absolutely not.

tests/e2e/fill-incorrect-identity-input-values.test.ts
```diff
@@ -48,11 +52,7 @@ describe("Fill incorrect identity input values", () => {
 
   describe("AddIdentity", () => {
     test("It should show error messages if Identity inputs are filled incorrectly in add identity", async () => {
-      expect.assertions(9);
```
Fenntasy, line 51: why?

Meyclem: Two reasons:
- For now, the function `authenticateToConnect` has some `expect` inside and we must add them in this count for the test to pass. I find that it make it useless because not reflecting the number of "real assertions" in the test.
- It's useful when something can pass or not, like with errors or calls or whatever that can "silently" fail, but here, we expect some things that will make the test fail if there are not there anyway. For example, we expect some input before trying to write inside. If the input is not there, the write action will fail and make the test fail in the same time.

Fenntasy: > It is an absolute hell to try our test suites with expect.assertions. And I'll find a better way to handle this but it's not for this PR

I'm curious on this point.

I would find it better to remove the expects from the `authenticateToConnect` since we don't need to test that though.

Meyclem: > It is an absolute hell to try our test suites with expect.assertions

Not talking about code here but we have many `expect`, and as I "play" with the tests and move things, I had to keep the count in mind and it was pretty annoying.

> I would find it better to remove the expects from the authenticateToConnect

I tried to not do that in this PR to keep changes  at a minimum because if I do, I have to change every `expect.assertions` for e2e tests.

As you prefer!

Fenntasy: > Not talking about code here but we have many expect, and as I "play" with the tests and move things, I had to keep the count in mind and it was pretty annoying.

You definition of "absolute hell" is quite a gentle one 😅 . Just comment the line while you're doing your tests and add it back at the end :)
But it's really a good practice to have it,I think it can save our beacon even in e2e tests.

> I tried to not do that in this PR to keep changes at a minimum because if I do, I have to change every expect.assertions for e2e tests. But can do, it's as you prefer!

Well you have my blessing 😇

Fenntasy (approved):

